PR #4, opened 2016-11-05
Support showing status bar, and in the problems section

orta:
screen shot 2016-11-05 at 16 13 43


orta:
doesn't do actual error line stuff yet, but let's not make perfect the enemy of done

PR #7, opened 2016-11-20
Look through the JSON for individual error messages

orta:
Now that 17 is in production, we can start doing error per line of code


orta:
vscode-jest

PR #14, opened 2016-11-29
Initial stab at Windows support

orta:
I don't have a windows machine around, so can't test this myself - @TanninOne and @onemen 

any chance you can update to 1.3.0 and confirm it's OK?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-### 1.3.0
+### 1.3.0 - 1.3.1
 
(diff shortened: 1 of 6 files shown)


TanninOne:
+  if (!path.includes(".bat") && platform() === "win32") {
+    return path + ".bat"; 
+  }

should be
+  if (!path.includes(".cmd") && platform() === "win32") {
+    return path + ".cmd"; 
+  }

Apart from that 1.3 seems to work.

PR #16, opened 2016-11-29
Handle fixing snapshots

orta:


src/extension.ts
```diff
@@ -220,99 +239,93 @@ class JestExt  {
         // OK - lets go
         this.parsingTestFile = true;
 
-        // try {
```
orta: just removed the indendation for this unused `try`

Issue #18, opened 2016-11-29
Handle on-boarding in a nicer way


gabro:
Ref: https://twitter.com/gabro27/status/803592079188365316

When installing the extension and opening an existing test file, all dots are white.
This is because jest only runs on changed files.

This is not a big deal, but the experience is not the most rewarding, as in "I installed this extensions to see green dots! Where are my green dots?!"

Possible ways of handling this:

- add a "run all" command that can be manually triggered
- make white dots empty green dots
- nothing, this is just me being nit-picky

Others?

orta:
Yeah, I think the unknown should be a green circle 👍

orta:
This is in

PR #19, opened 2016-11-29
Make not-ran tests indicator a green circle

gabro:
Issue #18 

Very simple change: a green circle is more rewarding than a lightgrey one.

I used the same green as the passed case.

Test case:

image


orta:
This looks good, can you add a line to the CHANGELOG to give yourself some credit? Then I'm not the only person on it 💃

orta:
Eh, I'll add it, I was going to wrap up the create-react stuff anyway :D

gabro:
Thanks! :)

PR #20, opened 2016-11-29
Initial work at adding support for a create-react app

orta:
Adds the ability for runner to work with different args when you're using a create-react app 👍 

screen shot 2016-11-29 at 18 53 27


orta:
Lightened the warning's text too 
screen shot 2016-11-30 at 06 58 19

PR #21, opened 2016-11-30
normalize path to make the default path to jest work on windows

bookman25:
I believe this should allow the default jestPath to work on Windows without having to override now.


orta:
Awesome, thanks @bookman25

PR #23, opened 2016-11-30
Migrate all non-vscode related code into a specific lib folder

orta:
Handles the:

>  Start to decouple VS Code from the codebase. This means only Extension.ts is allowed to use import {workspace} from 'vscode'; as a bare minimum.

part of this issue - https://github.com/facebook/jest/issues/2183

---

The `src/lib` directory is effectively self contained now, and does not have any references to VS Code. Making it available for Nuclide usage. 

/cc @cpojer

### DIFF
--- a/src/lib/types.ts
+++ b/src/lib/types.ts
@@ -1,0 +1,30 @@
+'use strict';
+
+export interface JestFileResults {
+    name: string;
+    summary: string;
+    message: string;
+    status: "failed" | "passed";
+    startTime:number;
+    endTime:number;
+    assertionResults: JestAssertionResults[];
+}
+
+export interface JestAssertionResults {
+    name: string;
+    title: string;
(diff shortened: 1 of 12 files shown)


cpojer:
*awesome*

src/lib/types.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export interface JestAssertionResults {
+    name: string;
+    title: string;
+    status: "failed" | "passed";
+    failureMessages: string[];
+}
+
+export interface JestTotalResults {
```
cpojer: The nice thing is that we have all those types in Jest already: https://github.com/facebook/jest/tree/master/types so once we merge there'll be some good de-duplication of this and it'll be easier to keep them in sync going forward!

orta: 👍

PR #24, opened 2016-11-30
fixes missing decorations and some 'this' reference bugs

KalleOtt:


orta:
hrm, this is a massive change because of some auto linters, I don't really know what has changed I'm afraid :/

orta:
also damn, 8 thumbs up in 3 hours :D

So this is hard, this project is in a transitory phase right now as I'm moving a lot of this code to facebook/jest - _and_ in the process converting it to Flow instead of typescript https://github.com/facebook/jest/pull/2192. Which I forked off this morning before these changes.

I'm going to keep this open, because I'll probably have to move to do the arrow syntaxing for the class functions too, and will need a reminder

KalleOtt:
sorry about the linter changes, happend automatically on save. 
I try to list my changes here:
1. the public class functions are now arrow functions so that the 'this' object is allways set to the object hte function was originally created on. There were some errors in the debugger because of wrong this-references
2. in extension..ts on line 224, the if statement was inverted, so that the decorator code was never on files matching the jest test pattern

orta:
Cool, so I've cloned the repo, and when I run this none of my tests get matched against the regex ->

screen shot 2016-11-30 at 13 00 42

I would need to change

screen shot 2016-11-30 at 13 02 02

to 

```js
    wouldJestRunURI = (uri: vscode.Uri): boolean => {
        const testRegex = new RegExp(this.jestSettings.settings.testRegex);
        const root = vscode.workspace.rootPath;
        const filePath = uri.fsPath;
        const relative = path.relative(root, filePath);
        return !!relative.match(testRegex);
    }
```

which re-inverts the logic again, could the regex be doing something I'm not expecting/ This is the first time I've used Regular expressions in JS ;)

KalleOtt:
I took a closer look into the wouldJestRunURI function and discovered that on a windows system the regex matching always failed because of the backslash path separator. 

A quick fix for now is to replace the backslashes in the path if the backslash is set as path system separator. 

And from the name of the function I think the return value should be true if the jest-test-regex gets a match, so I also changed the return statement.
With that change the negation in the if statement is again valid => leave the decorations function if the file is not matched by the jest regex

orta:
That explains it! 👍

orta:
Lets get this in, give it a run, and a release

orta:
I can just deal with the two source locations business for now

orta:
deployed as 1.6.1 👍

Issue #26, opened 2016-11-30
Support external config file


luizbon:
Version 1.6.1 is working with my environment, thanks @orta !

Since I'm using an external configuration file to run my jest tests, it fails. It is possible to add a new setting where I define the path to a config file?

I can try to do a PR with some guidance, never worked with VS Code extensions.

orta:
So the key will be to add a configuration to this:  https://github.com/orta/vscode-jest/blob/master/package.json#L55-L59

Then inside the [jest_process](https://github.com/orta/vscode-jest/blob/master/src/lib/jest_process.ts#L27) you need to add support for the extra arg `--config thingy.json` 

_however_ - because we have a separation of extension and library (due to https://github.com/facebook/jest/issues/2183 ) you cannot just plonk:

```js
  const jestSettings: any = workspace.getConfiguration("jest");
  const path = normalize(jestSettings.pathToJest);
```

in there, I'd recommend adding it a "default args" to the [workspace](https://github.com/orta/vscode-jest/blob/master/src/lib/project_workspace.ts) and pulling it out of that

PR #30, opened 2016-12-01
add config file option

luizbon:
proposed solution to issue #26

### DIFF
--- a/src/lib/jest_runner.ts
+++ b/src/lib/jest_runner.ts
@@ -23,10 +23,6 @@ export class JestRunner extends EventEmitter {
     start = () => {
         const tempJSON = tmpdir() + "/vscode-jest_runner.json";
         const args = ['--json', '--useStderr', '--watch', "--jsonOutputFile", tempJSON];
-        if (this.workspace.pathToConfig !== "") {
-            args.push("--config");
-            args.push(this.workspace.pathToConfig);
-        }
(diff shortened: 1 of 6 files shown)


src/lib/jest_runner.ts
```diff
@@ -23,6 +23,10 @@ export class JestRunner extends EventEmitter {
     start = () => {
         const tempJSON = tmpdir() + "/vscode-jest_runner.json";
         const args = ['--json', '--useStderr', '--watch', "--jsonOutputFile", tempJSON];
+        if (this.workspace.pathToConfig !== "") {
+            args.push("--config");
+            args.push(this.workspace.pathToConfig);
+        }
```
orta: you want this inside the jest_process instead, otherwise it won't happen on the other times we run Jest (generating configs, updating screenshots etc )

src/lib/project_workspace.ts
```diff
@@ -24,8 +24,14 @@ export class ProjectWorkspace {
    */
   pathToJest: string;
 
-  constructor(rootPath: string, pathToJest: string) {
+  /**
+   * The path to configuration file
```
orta: would love to write the semantics of how this is used too please ( e.g. "treat as ignored if zero length" )

luizbon:
Did the changes according to comments

orta:
Amaze - looks great 👍

orta:
deployed as 1.6.3 👍

PR #33, opened 2016-12-03
Fix the args array where config should be added

luizbon:
Found out that it was not working properly


orta:
hah - I understand, this stuff should get tests at some point - once I've moved it over to facebook/jest

PR #38, opened 2016-12-07
Fixed bug with create-react-app on Windows

brentatkins:
Fixes https://github.com/orta/vscode-jest/issues/17

- Use npm.cmd on windows
- Path comparison was not working


orta:
Perfect! This makes sense, so the first normalise was changing the route on windows but not the one used for comparison.

Nice debugging - Thanks @brentatkins

orta:
I'll make a release now

PR #48, opened 2016-12-20
jest-editor-support is 1 based for lines/columns, vscode is 0 based

bookman25:
Uses changes from https://github.com/facebook/jest/pull/2377


orta:
Alright, let's give this a shot - I worry that this will break the babylon stuff - as that was working previously

orta:
All looks good in practice 🔢

PR #50, opened 2016-12-21
break code into smaller modules

bookman25:
I was thinking we could start to split up and isolate parts of the code.  If we get around to writing some tests, it'll go easier with smaller modules. (fyi I tried to use jest to test, but because of the vscode dependencies I had some issues)

Let me know what you think.  If it's good I'll keep going, otherwise we can talk through next steps.

So far I've only pulled the JestExt class out into it's own file.


orta:
This looks 👍 -  totally fine with having this broken up

PR #54, opened 2016-12-22
clean before publish

bookman25:


orta:
Ah yeah! That's a lot of stuff to clean :D

I think this is cause I'm using `yarn link`

PR #55, opened 2016-12-22
wait for settings before executing

bookman25:
currently we trigger decorators if you start with an active editor.  it doesn't wait for loading the testRegex and as a result when you first open the project with a test file active it doesn't show any decorators.

this change loads the decorators after settings are loaded

### DIFF
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -14,6 +14,7 @@ import {
 } from 'jest-editor-support';
 
 import * as decorations from './decorations';
+import { IPluginSettings } from './IPluginSettings';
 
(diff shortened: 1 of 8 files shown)


src/JestExt.ts
```diff
@@ -128,23 +121,30 @@ export class JestExt {
         this.jestProcess.start();
     }
 
-    public stopProcess = () => {
+    public stopProcess() {
         this.channel.appendLine('Closing Jest jest_runner.');
         this.jestProcess.closeProcess();
     }
 
-    private getSettings = (completed: (settings: Settings) => void) => {
```
orta: nice one, this stuff was such hacks

PR #64, opened 2017-01-04
kill existing process before starting again

bookman25:
Extracted the statusbar updates into a separate file.
Terminate existing process before starting a new one.


orta:
Yes 👍 

Teset fail is legit though

PR #73, opened 2017-02-04
snapshot previews

bookman25:
Depends on snapshot support in jest-editor-support.  I'd like to do some more stuff with it, but I figure this is a good first start.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,6 +91,11 @@
           "description": "Whether codelens for debugging should show",
           "type": "boolean",
           "default": true
+        },
+        "jest.enableSnapshotPreviews": {
+          "description": "Whether snapshot previews should show",
+          "type": "boolean",
+          "default": true
         }
       }
     },
--- a/package.json
+++ b/package.json
@@ -114,8 +114,8 @@
     ]
   },
   "lint-staged": {
-    "*.json": ["npm run prettier-write --", "git add"],
-    "*.ts": ["npm run prettier-write --", "git add"]
+    "*.json": ["yarn prettier-write --", "git add"],
+    "*.ts": ["yarn prettier-write --", "git add"]
   },
   "scripts": {
     "precommit": "lint-staged",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -82,21 +82,21 @@
   "scripts": {
     "ci": "yarn lint && yarn test",
     "clean-out": "rimraf ./out",
-    "vscode:prepublish": "yarn clean-out && tsc -p ./src",
-    "compile": "tsc -watch -p ./src",
+    "vscode:prepublish": "yarn clean-out && tsc -p ./tsconfig.publish.json",
+    "compile": "tsc -watch -p ./tsconfig.publish.json",
```
orta: interesting, what were these changes for?

bookman25: Mainly because when the tsconfig was in a nested directory, I couldn't augment declarations.  So the extension of jest-editor-support.d.ts wasn't working properly with a nested tsconfig.

orta:
We're waiting on https://github.com/facebook/jest/pull/2629 incase people are wondering

bbigras:
facebook/jest#2629 was closed :(

Haroenv:
https://github.com/facebook/jest/pull/4570 has been merged :D

orta:
An alternative, as of the vscode out today, could be that we use a styled hover that shows the string as a colored diff

screen shot 2017-10-05 at 3 38 14 pm

bookman25:
Maybe I'm mistaken but it looks like completion items are for when you type things.  So you'd have to type out text in order to trigger the snapshot previews?

orta:
Ah, yeah, more specifically I was thinking about being able to maybe override the hover popup when you highlight the "toMatchSnapshot" with a preview

bookman25:
@orta anything still pending here?

orta:
If this is working (with the rebased PR), then nope 👍  should be good to go IMO

kizu:
I hope this would be finished and merged one day, as it is a thing I'd really want to see!

bookman25:
@orta I think this should be good now.  Rebased and appears to still be working locally :)

orta:
Alright, will give it a look through now

orta:
@bookman25 - this looks 👍 

screen shot 2017-12-29 at 12 19 00 pm

I'll happily merge and add this, can you add a way to disable it (as a setting) then this is good to go - congrats on bringing this PR back from the grave

package.json
```diff
@@ -114,8 +114,8 @@
     ]
   },
   "lint-staged": {
-    "*.json": ["npm run prettier-write --", "git add"],
-    "*.ts": ["npm run prettier-write --", "git add"]
+    "*.json": ["yarn prettier-write --", "git add"],
+    "*.ts": ["yarn prettier-write --", "git add"]
```
orta: 👍

PR #90, opened 2017-03-27
Fixes for create-react-app 

orta:
At some point either in the process of moving to jest-editor, or a CRA update, we lost the support for CRA. This mixed with https://github.com/facebook/jest/pull/3212 should fix CRA.

screen shot 2017-03-27 at 13 04 32


src/helpers.ts
```diff
@@ -18,7 +18,9 @@ export function pathToJest(pluginSettings: IPluginSettings) {
     const defaultCreateReactPathWindows = 'node_modules/react-scripts/node_modules/.bin/jest.cmd';
     const createReactPath = (platform() === 'win32') ? defaultCreateReactPathWindows : defaultCreateReactPath;
     const absolutePath = pluginSettings.rootPath + '/' + createReactPath;
-    if (!existsSync(path) && existsSync(absolutePath)) {
```
orta: I think something changed in CRA which includes jest in the `node_modules/.bin` now, or maybe I was just doing it wrong from the get-go. Now we just look for the CRA version outright and assume that must make it a CRA.

PR #91, opened 2017-03-28
Add back our TypeScript parser

orta:
In https://github.com/facebook/jest/pull/2973 the TypeScript parser was moved into another node module, so we needed to make the choice in which parser we would using inside this extension.

In updating to the latest jest-editor, we needed this or we go no TypeScript support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,6 +105,7 @@
   },
   "dependencies": {
     "elegant-spinner": "^1.0.1",
-    "jest-editor-support": "18.1.0"
+    "jest-editor-support": "19.0.2",
+    "jest-test-typescript-parser": "18.1.0"
   }
 }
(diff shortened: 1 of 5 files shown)


orta:
Note, that I can only deploy from a `yarn link`ed `jest` until the next release of jest due to all this

package.json
```diff
@@ -105,6 +105,7 @@
   },
   "dependencies": {
     "elegant-spinner": "^1.0.1",
-    "jest-editor-support": "18.1.0"
+    "jest-editor-support": "19.0.2",
+    "jest-test-typescript-parser": "18.1.0"
```
bookman25: Why not the latest version of the parser?

yarn.lock
```diff
@@ -1660,6 +1654,13 @@ jest-snapshot@^18.0.0:
     natural-compare "^1.4.0"
     pretty-format "^18.0.0"
 
+jest-test-typescript-parser@18.1.0:
+  version "19.0.2"
```
bookman25: lol, looks like you got the latest version of the parser, but package.json doesn't have it.

PR #131, opened 2017-07-29
Apply prettier to the codebase

macklinu:
https://github.com/prettier/prettier

I was poking around this morning, and when I save files in my IDE, prettier formats and writes that file. If you're open to it, I've run prettier using the configuration found in danger-js. This PR also sets up running prettier on staged files using lint-staged.

Lemme know what you think. If you're like the prettier configuration changed to include semicolons or anything else, happy to make those changes. 👍


orta:
👍

orta:
Perfect

Issue #98, opened 2017-04-24
Exception raised: Process failed: spawn node_modules\.bin\jest.cmd ENOENT


bnwan:
Enable to get this working. Using create-react-app with TypeScript. Any help appreciated

orta:
Can you provide a demo project?

iamfreee:
Getting this too.

Windows 10, clean create-react-app project:
![image](https://cloud.githubusercontent.com/assets/1500881/25660430/11e857fa-3004-11e7-8826-abe2ea770ac1.png)

3KniGHtcZ:
I have same problem on macOS

BigForNothing:
Same problems, windows 7 & windows 2012r2. Also got an error about needing to run at least jest 20+ and I'm running 21.2.1

Fendrian:
I'm not using create-react-app, but I am using Windows 10 and struggled with that precise error message with my project for some time and have finally managed to find a functional workaround.

The basic issue seems to be with the path finding code on Windows. As a result of the fix to #10, when VSCode is running in a Windows environment this extension naively assumes that everything must end in .cmd, and actually adds .cmd to the end of `jest.pathToJest` if it doesn't already have it. This may be reasonable functionality in some specific use cases, but sadly is completely wrong in my case - I may be developing on Windows 10, but literally nothing in my project folder ends in .cmd.

In the end the only way I could find to satisfy the requirements was to create a text file called `jest.cmd` in my project root, and make use of the `%*` syntax:

`node node_modules\jest-cli\bin\jest.js %*`

The `%*` passes any arguments provided to the .cmd file to the inner script so the suffixes and arguments all work, and because the script is inside a file that ends in .cmd I can actually point the path at it, so everything works!

I hope it works for others too.

akaztp:
Or you can install jest globally and set the config:
```
"jest.pathToJest": "jest"
```

seanpoulter:
Is this one still a problem? Could anyone experiencing the issue provide more info to reproduce the steps or create a sample repo?

Fendrian:
I haven't had time to dig into this deeply, but yes, it's still an issue. Skimming the source code, it looks like the issue is still lines 22-24 of `vscode-jest/src/helpers.ts` still naively sticking `.cmd` on the end of `jest.pathToJest` in Windows environments.

[...]

PR #137, opened 2017-08-23
adding coverage overlay

bookman25:


orta:
This looks 👍 - not sure what's up with those tests though

bookman25:
All green now 😄

orta:
amaze

src/Coverage/CoverageCodeLensProvider.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export function registerCoverageCodeLens(jestExt: JestExt) {
+  return [
+    vscode.languages.registerCodeLensProvider(
+      { pattern: '**/*.{ts,tsx,js,jsx}' },
+      new CoverageCodeLensProvider(jestExt)
+    ),
+    vscode.commands.registerCommand(coverageCommand, () => {
+      vscode.window.showInformationMessage('You have unit tests!')
+    }),
```
seanpoulter: Were we planning to use this command for anything? It looks like we could remove it now.

bookman25: At one point I had a plan for it.  Not clue what it was now though :)  I think I was planning on maybe using it to toggle the coverage overlay (ended up doing that quick/dirty later though).

seanpoulter: Oh, I know that feeling. I'll pop that out now.

stephtr:
Is there any information or sample available on how to enable showing the coverage information?
I tried it for my own project, vscode-jest and create-react-app within vscode-jest, but without success.

seanpoulter:
We don't have an example ready to go @stephtr but we'd appreciate a PR to improve the documentation or add to the example in `integrations/create-react-example`.
<br/>

You might find this repo I've been using to troubleshoot Issues helpful: [seanpoulter/vscode-jest-181](https://github.com/seanpoulter/vscode-jest-181). You'll see that Jest is configured to collect coverage in `jest.config.js`, and the setting is enabled in `.vscode/settings.json` to show the coverage when the test results come in.

stephtr:
Thank you!

In order to enable the coverage informations when using create-react-app, one has to manually add `jest.pathToJest` to include `--coverage`.
Wouldn't it be more useful to add a new vscode setting (a la "collectCoverageByDefault") which automatically appends the `--coverage` flag? An advantage would be that one doesn't have to specify for Windows or Linux/Mac by including .cmd (`npm.cmd`) or not. That new setting would also work for other projects directly running jest (instead of `npm test`).

seanpoulter:
> In order to enable the coverage informations when using create-react-app, one has to manually add jest.pathToJest to include --coverage.

You can also add it to the `scripts` in `package.json`. Since collecting coverage adds a performance hit  I've seen quite a few variations of `test` and `test-coverage` scripts.
<br/>

> Wouldn't it be more useful to add a new vscode setting (a la "collectCoverageByDefault") which automatically appends the `--coverage` flag?

That's an interesting idea. We welcome PRs if you're keen. For my day to day needs, I'd hold off using that setting until we could run tests related to the open files (#205) and improve the coverage overlay a bit (#177, local WIP).
<br/>

> An advantage would be that one doesn't have to specify for Windows or Linux/Mac by including .cmd (`npm.cmd`) or not.

Hrm. Does that mean a repo with `pathToJest` set wouldn't work across platforms? I think we'll need to fix this one at the root cause, and clean up the other issues related to running the right binaries: [#98](https://github.com/jest-community/vscode-jest/issues/98#issuecomment-353717341), #193.

stephtr:
> You can also add it to the scripts in package.json. Since collecting coverage adds a performance hit I've seen quite a few variations of test and test-coverage scripts.

My thought was to have an option for enabling coverage overlay without having to edit scripts in package.json such that one could still manually run `npm test` without generating/displaying the generated coverage information (even though there is still the possibility to set `"jest": {"coverageReporters": []}` in package.json).

> That's an interesting idea. We welcome PRs if you're keen. For my day to day needs, I'd hold off using that setting until we could run tests related to the open files (#205) and improve the coverage overlay a bit (#177, local WIP).

The target audience in my mind were users with sparse experience who just run `create-react-app` (small projects). For them it would be more self-explaining and discoverable to toggle one VS Code setting than adding a script flag in package.json.

> Hrm. Does that mean a repo with pathToJest set wouldn't work across platforms? I think we'll need to fix this one at the root cause, and clean up the other issues related to running the right binaries: #98, #193.

vscode-jest sticks a `.cmd` to the end of a custom `jest.pathToJest` if not already present, which is not quite useful for `"jest.pathToJest": "npm test -- --coverage"` (or running with other flags). In order to prevent `Process failed: spawn npm ENOENT` on therefore has to write `"npm.cmd test -- --coverage"`, which isn't portable across different plattforms.

seanpoulter:
I should've suggested this before -- let's move future discussion to the issues and give the poor CCed folks a rest.

PR #140, opened 2017-08-25
display error messages like failed assertions inline

### LINKED ISSUE #58: Doesn't display JS error message,
First; this is a very cool extension.

When you get a JS error you'd expect it to display as thus:

```js
state = STORE.getStateXXX() // TypeError: STORE.getStateXXX is not a function
```

Instead it's 

```js
state = STORE.getStateXXX() //
```
Still useful, but could use a message.

bookman25:
fixes #58


orta:
👍 Makes sense yeah

PR #145, opened 2017-09-30
chore(settings): allow settings autocompletion

vvo:
image

I don't really know why but I got "unknown" setting when typing "jest.pathToJest" in vscode. While with other extensions like flow it works perfectly.

It seems you are missing a `"type": "Object"` prop and also have some extra properties like titles for every option. Compared to flow that's the only difference.

vscode doc also says we may want to add type information (https://code.visualstudio.com/docs/extensionAPI/extension-points#_contributesconfiguration).

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,22 +42,21 @@
       }
     ],
     "configuration": {
+      "type": "object",
+      "title": "Jest configuration",
       "properties": {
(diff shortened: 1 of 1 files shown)


orta:
Great - awesome 👍

Issue #129, opened 2017-07-19
Support monorepos


Pajn:
It would be great if this plugin supported monorepos. Either by passing the `--projects` option to jest or by starting a jest process for every subproject (folder with `package.json` file in it). In the case of multiple Jest processes the current setting paths should be relative from each subproject.

orta:
As you can customise the command used to trigger jest (`jest.pathToJest`) then I think specific support for projects probably doesn't need to go in

Pajn:
I have tried adding a script to the root package.json with all projects specified with `--projects` and specifying `yarn test --` as `jest.pathToJest` but then there are no changes detected even though the same command in the terminal detects changes an run just fine.

ls-guillaume-lambert:
It does support monorepos, we have it working here for one of our projects using lerna. Jest resides at the root of your monorepo, correct? We added this to our `.vscode/settings.json` file:

```
"jest.pathToConfig": "./jest.config.js",
"jest.pathToJest": "./node_modules/.bin/jest",
```

connectdotz:
we also have a lot of problems with vscode-jest for our lerna managed monorepo.

we would like to have the root vscode workspace able to watch for all file changes in the packages, report test failures on the problems inspector and update test decorators just like in the single project case.

We have tried both approaches:
1. use jest's monoreport support, i.e. use the `projects` in jest configuration.
    * problem 1: during "--showConfig", jest report different config file format that includes all package's config into "configs" (instead of "config" for single project), therefore vscode-jest failed to parse the config output and will fallback to the default settings, which currently only looking for js(x) pattern.
   * problem 2: this is not vscode-jest issue, but jest didn't seem to be able to run `jest --watch` for monoreport, it got stuck at the first package, so this approach might simply not feasible at this point...
1. use lerna's test support, i.e. `"jest.pathToJest": "lerna run --parallel test --",`
   * problem 1: there are many config file chunks reported back during the `--showConfig`, which vscode-jest didn't seem to know how to handle, so again it fallback to the default settings. 
    * problem 2: the jest runner used the --outputFile with a tmp file when running the tests, lerna will just copy this command to every package, therefore they keep override each other's test result. 

To further complicate the case, our packages are not uniformed, i.e. there are typescript, javascript, react and react-native, therefore the default settings, which only look for js(x) patterns, doesn't work for us.

Maybe we are missing something obvious, @ls-guillaume-lambert, do you mind sharing with us how you get it to work? Thx.

[...]

PR #146, opened 2017-10-01
improve failure reporting for the whole project

connectdotz:
This PR addressed the issues discussed in #129, mainly to make managing tests easier (especially for monorepo projects) by providing the test state for the "whole-project" in addition to the "changed", plus a few usability/stability enhancement.

**change outline**
* added an initial full test run (a.k.a. non-watch-mode) phase when the runner first started, followed by the watch-mode run so the subsequent changes can be tested efficiently without losing sight of the whole project state. 
* all the failed tests will be shown in the problem inspectors independent of the editors.
    * note, consequently, only the line position is available in the inspector.
* fixed an issue that, under some race condition, caused the spinner to spin forever.
* added more information on the status bar, such as watch-mode/initial run, number of failed test suites etc.
* upgraded almost all the dependencies to the latest versions
* some refactoring to improve efficiency, stability, and readability. 
* improve test coverage

**Note**
* This PR depends on jest [#4572](https://github.com/facebook/jest/pull/4572).

**Follow up**
* it might be helpful to document the limit of vscode-jest monorepo support, such as it doesn't support lerna style package-level-testing, i.e. multiple jest processes/output; and that it requires one jest config for the whole project...
* large project might take a bit longer for the initial full test run, wondering if we should add an option to turn this feature on/off. If we do allow "partial" test reporting, we should clearly state it in the statusBar to avoid confusion/misleading.

### DIFF
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -288,7 +288,7 @@ export class JestExt {
 
   private resetInlineErrorDecorators(editor: vscode.TextEditor) {
     this.failingAssertionDecorators.forEach(element => {
-      editor.setDecorations(element, [])
+      element.dispose()
     })
     this.failingAssertionDecorators = []
   }
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -1,7 +1,6 @@
 import * as vscode from 'vscode'
 import * as path from 'path'
 import {
-  Expect,
   ItBlock,
   Runner,
   Settings,
@@ -10,6 +9,7 @@ import {
   TestReconciler,
   JestTotalResults,
   IParseResults,
+  TestAssertionStatus,
 } from 'jest-editor-support'
 import { parse as typescriptParse } from 'jest-test-typescript-parser'
 import { matcher } from 'micromatch'
@@ -21,6 +21,7 @@ import { TestReconciliationState } from './TestReconciliationState'
 import { pathToJestPackageJSON } from './helpers'
 import { readFileSync } from 'fs'
 import { Coverage, showCoverageOverlay } from './Coverage'
+import { updateDiagnositics, resetDiagnositics, failedSuiteCount } from './diagnostics'
 
 export class JestExt {
   private workspace: ProjectWorkspace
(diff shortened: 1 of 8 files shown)


src/JestExt.ts
```diff
@@ -354,20 +362,27 @@ export class JestExt {
...
-    if (data.success) {
+    const results = this.reconciler.updateFileWithJestStatus(data)
+    updateDiagnositics(results, this.failDiagnostics)
+
+    const failedFileCount = failedSuiteCount(this.failDiagnostics)
+    if (failedFileCount <= 0 && data.success) {
       status.success()
     } else {
-      status.failed()
+      status.failed(` (${failedFileCount} test suite${failedFileCount > 1 ? 's' : ''} failed)`)
```
orta: nice move 👍

src/diagnostics.ts
```diff
@@ -0,0 +1,62 @@
+/**
+ * this module contains functions to show jest test results in
+ * vscode inspector via the DiagnosticsCollection.
+ */
+import * as vscode from 'vscode'
+// import { DiagnosticCollection, Uri, Diagnostic, Range, DiagnosticSeverity } from 'vscode'
+import { TestFileAssertionStatus } from 'jest-editor-support'
+import { TestReconciliationState } from './TestReconciliationState'
+
+export function updateDiagnositics(testResults: TestFileAssertionStatus[], diagnostics: vscode.DiagnosticCollection) {
```
orta: `updateDiagnositics`, typo: `updateDiagnostics`

src/diagnostics.ts
```diff
@@ -0,0 +1,62 @@
...
+  })
+}
+
+export function resetDiagnositics(diagnostics: vscode.DiagnosticCollection) {
+  diagnostics.clear()
+}
+export function failedSuiteCount(diagnostics: vscode.DiagnosticCollection): number {
+  let sum = 0
+  diagnostics.forEach(() => sum++)
+  return sum
```
orta: interesting, I didn't know `reduce` was not a part of the JS core library

src/statusBar.ts
```diff
@@ -20,27 +20,27 @@ export function initial() {
...
 
-export function stopped() {
-  updateStatus('stopped')
+export function stopped(details?: string) {
+  updateStatus('stopped', details)
   setTimeout(() => initial(), 2000)
 }
 
-function updateStatus(message: string) {
+function updateStatus(message: string, details?: string) {
```
orta: nice move

cpojer:
Published 21.3.0-alpha.eff7a1cf for you guys.

orta:
ohhh yeahhh, thanks!

src/diagnostics.ts
connectdotz: Array.prototype.reduce() is indeed part of JS core, unfortunately diagnostics is a `vscode.DiagnosticCollection`, which only exposes forEach() iterator... I know it looked stupid but we have no choice here...

src/diagnostics.ts
connectdotz: Oops, indeed... Looks like you already merged it, I will submit a new PR then

src/diagnostics.ts
orta: 👍

PR #154, opened 2017-10-11
Add support for using sub-folder as jest root path

garyxuehong:
One scenario is: The project is a maven java project. Frontend code and
package.json/.jestrc files are in sub folder. Need to add a new config
'rootPath' to support run jest command in the subfolder

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,11 @@
           "type": "string",
           "default": ""
         },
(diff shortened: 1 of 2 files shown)


connectdotz:
is this duplicating jest's [rootDir](https://facebook.github.io/jest/docs/en/configuration.html#rootdir-string) config?

PR #163, opened 2017-11-04
Added code lenses for running and debugging tests (draft)

CzBuCHi:
**Code is not ready to be merged.** Nothing is broken, but now there are two ways how to run tests and i need your input how to merge them ... 

After polishing this i will create new fork with only one 'correct' commit and make PR from that.

**my changes**:

- added dependency on my unpublished package [vscode-test-runner](https://github.com/CzBuCHi/vscode-test-runner) - handles vscode integration (lenses, commands) without any knowelege about jest. Cannot publish it until mid of next week...
- added configuration properties `babylonPlugins`, `debugPort`, `debugTrace`, `globPatterns`, `jestCli`, `nodePath`, `stateJsonPath`, `tsConfigPath` and `webpackConfigPath`
- my code can handle webpack compilation, but yours shows syntax errors, with is confusing - open [this example project](https://github.com/CzBuCHi/vscode-test-runner/tree/master/examples/example-jest-webpack) in extension host instance of vscode to see that i mean ...
  - this would be fixed by using my version of test runner, but that will break backward compatibility ...
- my code reads directly from `WorkspaceConfiguration` with makes it non-agnostic
- my code needs path to `node_modules/jest-cli/bin/jest.js`, but yours is using `node_modules/.bin/jest`
  - i could parse  `node_modules/.bin/jest` and extract path to `jest.js`, but that would break if `npm test --` is used ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,5 @@
 # vscode-jest [![Build Status](https://travis-ci.org/orta/vscode-jest.svg?branch=master)](https://travis-ci.org/orta/vscode-jest)
 
-## Fork changes
-- added dependency on unpublished package [vscode-test-runner](https://github.com/CzBuCHi/vscode-test-runner) (cannot publish now), needs to be `npm link`ed 
-- added configuration properties `babylonPlugins`, `debugPort`, `debugTrace`, `globPatterns`, `jestCli`, `nodePath`, `stateJsonPath`, `tsConfigPath` and `webpackConfigPath`
-  - note: `jestCli` added because my code needs path to `node_modules/jest-cli/bin/jest.js`
-- my code can handle webpack compilation, but yours shows bunch of syntax errors - see [this example project](https://github.com/CzBuCHi/vscode-test-runner/tree/master/examples/example-jest-webpack)
-- my code reads directly from `WorkspaceConfiguration` with makes it non-agnostic
-
 ## The Aim
 
 A comprehensive experience when using [Facebook's Jest](https://github.com/facebook/jest) within a project.


orta:
This is amazing, sorry if you feel like I guilted you into it :D

CzBuCHi:
meh ... better that having zillion extension installed for same thing ... anyway di u look at it and how to merge jest runners?

orta:
> my code needs path to node_modules/jest-cli/bin/jest.js, but yours is using node_modules/.bin/jest

The problem here is the extra places that jest could be referenced in ( [e.g. CRA](https://github.com/orta/vscode-jest/blob/master/src/helpers.ts#L28)) 

---

WRT the webpack example, I'm not sure, I've never worked on a web JS project ( I mostly focus on CLI tools and React Native ) - this project has always just used the principal of "start jest in watch mode, and listen for the JSON" which should ideally work for any sort of Jest setup

orta:
I'm willing to switch the entire runner to your version though, assuming it works for most projects (and handles the CRA/CRNA edge cases )👍

CzBuCHi:
i will try to incorporate CRA in my VTR package and the go back to u ... BTW: jut got WTF is WRT moment :D

orta:
LOL

CzBuCHi:
Finally got some time for this ... VTR dont care witch test framework is used under the hood - it only provides lenses and commands for vscode and calls simple callback witch is running test - so CRA needs to be resolved here.

Jest watch mode will work, if results are saved to json (see `FileStates` type in VRT typings.d.ts) - VTR will watch that file and updates lenses accordingly.

Replacing runner is basically wrapping it in method with this signature:
```ts
export type executeTestCallback = (
  channel: vscode.OutputChannel,
  fileName?: string,
  testNamePattern?: string,
  debug?: boolean,
) => Promise<void>;
```
and send it as callback to VTR.
Currently im fixing bug in VTR (tsx not parsed correctly) so it will take me some time ...

CzBuCHi:
created new PR with cleaner diff with less code :)

Issue #159, opened 2017-10-26
"Illegal argument: line must be positive" error thrown when testing an Angular service with missing provider


asgallant:
### Environment

1. `node -v`: 8.6.0
2. `npm -v`: 5.3.0
4. `npm ls react-scripts` (if you haven’t ejected): n/a

1. Operating system: Windows 7

Jest version 21.2.1

### Steps to Reproduce

Create a service:
```typescript
import { Injectable } from '@angular/core';
@Injectable()
export class FooService {}
```
Write a test, missing a provider for FooService:
```typescript
import { inject, TestBed } from '@angular/core/testing';

import { FooService } from './foo.service';

describe('FooService', () => {
  beforeEach(() => {
    TestBed.configureTestingModule({
      imports: [],
      providers: []
    });
  });

  it('should be created', inject([FooService], (service: FooService) => {
    expect(service).toBeTruthy();
  }));
});
```

### Expected Behavior

Test suite fails with error "No provider for FooService" (this is what the command-line Jest does).

### Actual Behavior

vscode-jest throws an exception:
```
[Orta.vscode-jest] Illegal argument: line must be positive: Error: Illegal argument: line must be positive
	at Object.t.illegalArgument (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\node\extensionHostProcess.js:28:56657)
	at new e (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\node\extensionHostProcess.js:28:237802)
	at new e (c:\Program Files\Microsoft VS Code\resources\app\out\vs\workbench\node\extensionHostProcess.js:28:240093)
	at diagnostics.set.asserts.map.assertion (C:\Users\dgallant\.vscode\extensions\Orta.vscode-jest-2.4.2\out\src\diagnostics.js:20:48)
	at Array.map (native)
	at addTestsError (C:\Users\dgallant\.vscode\extensions\Orta.vscode-jest-2.4.2\out\src\diagnostics.js:18:38)
	at testResults.forEach.result (C:\Users\dgallant\.vscode\extensions\Orta.vscode-jest-2.4.2\out\src\diagnostics.js:33:21)
	at Array.forEach (native)
	at Object.updateDiagnostics (C:\Users\dgallant\.vscode\extensions\Orta.vscode-jest-2.4.2\out\src\diagnostics.js:25:17)
[...]

Issue #156, opened 2017-10-16
Illegal argument: line must be positive: Error: Illegal argument: line must be positive


capaj:
### Environment

1. `node -v`: 8.6.0
2. `npm -v`: 5.4.1
4. `npm ls react-scripts` (if you haven’t ejected): [fill]

1. Operating system: ubuntu 17.04

### Steps to Reproduce

opened up my project and saved a file

### Expected Behavior

no error is thrown

### Actual Behavior

error is thrown:
```

/usr/share/code/resources/app/out/vs/workbench/workbench.main.js:29 [Orta.vscode-jest] Illegal argument: line must be positive: Error: Illegal argument: line must be positive
	at Object.t.illegalArgument (/usr/share/code/resources/app/out/vs/workbench/node/extensionHostProcess.js:28:56657)
	at new e (/usr/share/code/resources/app/out/vs/workbench/node/extensionHostProcess.js:28:237724)
	at new e (/usr/share/code/resources/app/out/vs/workbench/node/extensionHostProcess.js:28:240015)
	at diagnostics.set.asserts.map.assertion (/home/capaj/.vscode/extensions/Orta.vscode-jest-2.4.2/out/src/diagnostics.js:19:48)
	at Array.map (native)
	at addTestsError (/home/capaj/.vscode/extensions/Orta.vscode-jest-2.4.2/out/src/diagnostics.js:17:38)
	at testResults.forEach.result (/home/capaj/.vscode/extensions/Orta.vscode-jest-2.4.2/out/src/diagnostics.js:32:21)
	at Array.forEach (native)
	at Object.updateDiagnostics (/home/capaj/.vscode/extensions/Orta.vscode-jest-2.4.2/out/src/diagnostics.js:24:17)
	at JestExt.updateWithData (/home/capaj/.vscode/extensions/Orta.vscode-jest-2.4.2/out/src/JestExt.js:307:23)
	at Runner.__dirname.startProcess.jestProcess.on.on (/home/capaj/.vscode/extensions/Orta.vscode-jest-2.4.2/out/src/JestExt.js:61:18)
	at emitOne (events.js:96:13)
	at Runner.emit (events.js:191:7)
	at __dirname.default.start.debugprocess.stdout.on.data (/home/capaj/.vscode/extensions/Orta.vscode-jest-2.4.2/node_modules/jest-editor-support/build/Runner.js:67:18)
	at tryToString (fs.js:449:3)
[...]

PR #167, opened 2017-11-10
addressing illegal argument error when passing invalid line number

connectdotz:
this is to address the symptom in #159 and #156. 

Note:
1. this doesn't address the actual root cause (that jest error parsing yield unexpected result), but merely making a best effort attempt to show the error as received. Hopefully, it will provide more hint on what the actual underlying error is...
1. I didn't have an actual use case to test against, only unit tested...


orta:
Yeah, this makes sense 👍

PR #169, opened 2017-11-15
assertionsForTestFile can return null

BBB:
So let's default to []

https://github.com/facebook/jest/blob/38c2cf7d0c31300aa08034fc3903f615f7bd7825/packages/jest-editor-support/src/test_reconciler.js#L147


orta:
Agree 👍

orta:
Thanks

Issue #173, opened 2017-11-18
Lingering failures after file deletions in diagnostics / problems pane


erikns:
### Environment

1. `node -v`: v8.4.0
2. `npm -v`: v3.7.3
3. Operating system: macOS 10.12.6

### Steps to Reproduce

1. Start a Jest watcher session in VSCode. The initial run completes and passes.
2. Add a new test file that has a failure
3. The failure shows up in the status bar and problems pane as expected
4. Delete the new failing file

### Expected Behavior

One should expect the failure to go away when it is no longer relevant, such as when files are removed.

### Actual Behavior

The failing file remains in the status bar and problems pane.

![image](https://user-images.githubusercontent.com/11320269/32985093-ae3b5274-ccb2-11e7-974f-8b1cc17e1a2f.png)

### Fix

I have created a patch that fixes this issue locally. It is pretty simple. All it does is go through the existing diagnostics, and removes the ones that correspond to files that no longer exist. I would be happy to submit a PR for this.

orta:
Sure, that sounds 👍

PR #174, opened 2017-11-18
Remove diagnostics for deleted files

erikns:
PR fixing issue #173 

It runs through the diagnostics every time they are updated, searching for files that are no longer there, and then removes them.


orta:
Rockin'

orta:
This is shipped in 2.5.0

Issue #116, opened 2017-06-01
dot doesn't update on format a document without save


tiagor87:
When you run Jest, it put some passed and failed dots in tests, but if you have to format the text, the dots keeps in the middle of the text. Not a big problem at all, but I think we should improve the tool.

orta:
Yeah, sure, I think that's about using the document edit API from viscose to update the dots - it's just a matter of adding it, I've been hesitant because it feels like a lot of code to be running all the time but maybe if we had a heuristic for updating them like total number of lines of code has changed?

tiagor87:
Yes @orta, I will try to develop this feature and do some tests on Saturday.

seanpoulter:
Does this work the way you expect it to now? Seems like they update position (but not the state based on the unsaved document contents).

seanpoulter:
Looks like we can improve this quite a bit. The dots are displayed **every time** the document changes, and not configured to "stick" to the test block. Here's an annotated "call graph" that walks through how the dots are unnecessarily applied every time the file changes.

```js
// extension.ts
activate(context) {
  context.subscriptions.push(
    ...registerFileChangeWatchers(instance)
  )
}

// fileChangeWatchers.ts
registerFileChangeWatchers(extension) {
  vscode.workspace.onDidChangeTextDocument({ contentChanges, document }) => {
    extension.triggerUpdateDecorations(editor)
  })
}

// JestExt.ts
triggerUpdateDecorations(editor) {
  showCoverageOverlay(editor, this.coverage)
  // The coverage would have changed, but we haven't run the tests with the unsaved buffer

  if (!canUpdateDecorators(editor)) { ... }
  // Only changes if the Jest settings for testMatch or testRegex change

  updateDotDecorators(editor)
  // The dots may change, but we haven't run the tests with the unsaved buffer
}

// Coverage/overlay.ts
showCoverageOverlay(editor) {
  // Decorators applied using stale coverage. 
  // These decorators don't have a rangeBehavior set.
}

// JestExt.ts
canUpdateDecorators(editor) {
  wouldJestRunURI(uri)
}

wouldJestRunURI(
  // Return value only changes when Jest settings for testMatch or testRegex change
}

updateDotDecorations() {
  // The sorted test results may change, but we haven't run tests with the unsaved buffer.
  // Until then the sorted test results, dot decorations, Debug CodeLens, and inline errors don't change.

  // The dot decorators and inline errors don't have the rangeBehavior set.
[...]

PR #200, opened 2017-12-25
Improve the dot decoration placement as we edit

seanpoulter:
The dot decorations would often appear in the wrong spot as we edit the document because they updated after every change using locations from the last results.   
<br/>

This commit stops updating the test result decorations when there are unsaved changes to the document. Using the `onDidChangeTextDocument` event handler, we can detect when the document is saved or updated by an external editor and update accordingly. It also is fired when other Resources are changed (e.g.: using the `git` scheme from the Git extension).

To keep the decorations in place as we edit or format the document, the the decoration types are now created with a `DecorationRangeBehavior` that prevents the range from growing and gives them a "sticky" feel.

This commit also changes the way the inline error messages are cached to keep the error messages visible in all instances of a document, not just in the active editor.
<br/>

This addresses most of #116 but there are at least two problems:
* the Debug CodeLens is in the wrong spot, and the API doesn't seem to support range "stickiness"
* if you open the editor to a document with unsaved changes, the dots appear in the wrong spot

Both of those would be solved with a language service that keeps track of the test positions and results for unsaved changes.


src/JestExt.ts
```diff
@@ -545,4 +552,47 @@ export class JestExt {
...
     this.testResultProvider.removeCachedResults(filePath)
   }
+
+  removeCachedDecorationTypes(document: vscode.TextDocument) {
+    if (!document || !document.fileName) {
+      return
+    }
+
+    delete this.failingAssertionDecorators[document.fileName]
+  }
```
orta: nice last week I almost made a PR for all this - https://github.com/jest-community/vscode-jest/tree/turn_off_stuff

orta: well, the removal anyway

src/decorations.ts
```diff
@@ -16,6 +16,7 @@ export function failingItName() {
         contentText: '●',
       },
     },
+    rangeBehavior: DecorationRangeBehavior.ClosedClosed,
```
orta: so closed they wrote it twice

tests/extension.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    it('should register an event handler to handle when a document is saved', () => {
+      const handler = jestInstance.onDidChangeTextDocument
+      activate(context)
+
+      expect(vscode.workspace.onDidChangeTextDocument).toBeCalledWith(handler, thisArg)
+      expect(context.subscriptions.push).toBeCalled()
+      expect(context.subscriptions.push.mock.calls[0]).toContain('onDidChangeTextDocument')
+    })
+  })
+})
```
orta: damn! This is a lot of tests, congrats

orta:
TBH, converting jest-editor-support to be a language server would be an epic move

orta:
Shipped as 2.5.6 👍 

Nice work, we should get you on the README somewhere as a maintainer

src/decorations.ts
seanpoulter: It's a terrible name, and unfortunately they haven't added the legible [alias for it](https://github.com/Microsoft/vscode/blob/114adf703be6203fba2da696aeac439449241ac9/src/vs/editor/common/editorCommon.ts#L899) to the enum:
```
export enum TrackedRangeStickiness {
	AlwaysGrowsWhenTypingAtEdges = 0,
	NeverGrowsWhenTypingAtEdges = 1,
	GrowsOnlyWhenTypingBefore = 2,
	GrowsOnlyWhenTypingAfter = 3,
}
```

The default is `OpenOpen` or `AlwaysGrowsWhenTypingAtEdges`.

src/JestExt.ts
seanpoulter: Nice. Assign me to the review if you get **that** annoyed with them again. :wink:

seanpoulter:
I've created Issue #205 to discuss the language server as we need to @orta.

PR #207, opened 2017-12-27
Fix dot decoration display on Windows

### LINKED ISSUE #206: dot decorations not working properly
### Environment

1. `node -v`: 8.9.1
2. `npm -v`: 5.6.0
3. Operating system: Windows Server 2016

### Issue

The dot decorations aren't working correctly after upgrading to 2.5.7. Inline failure messages also don't seem to be working.

### Expected Behavior (using 2.5.6)

image

image

### Actual Behavior (using 2.5.7)

image

image

seanpoulter:
The test result lookup failed on Windows **again** after #204. This change moves the adapter into `updateResults`. The source of the problem is the difference between:
* the TextDocument URI always having a lowercase drive letter
* the Jest test results having a drive letter that is system dependent
I incorrectly assumed Jest was sending us uppercase drive letters for #204, sorry. :hankey: :disappointed:

Confirmed this looks OK from my Linux and Windows boxes.

This closes #206.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 
 ## Master
 
-### 2.5.7
+*
+
+### 2.5.7-8
 
 * Fix dot decoration display on Windows - seanpoulter
 
--- a/src/TestResultProvider.ts
+++ b/src/TestResultProvider.ts
@@ -53,7 +53,7 @@ export class TestResultProvider {
     }
 
     const { itBlocks } = parseTest(filePath)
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,9 @@
 
 ## Master
 
-### 2.5.7
+*
+
+### 2.5.7-8
```
orta: Heh, we've all been there

PR #208, opened 2017-12-28
Hide the notes and work in progress

seanpoulter:
When looking at the extension Changelog in VS Code the notes and Master are visible but not relevant. I'm feeling like this is #trivial and shouldn't go in the changelog.

image


orta:
Good thinking

PR #215, opened 2018-01-04
Clean up jest process management

marcinczenko:
I added two basic abstractions to make jest process management easier to managed: `JestProcess` and `JestProcessManager`. `JestProcess` job is to isolate the rest of the extension from the underlying `jest-editor-support`. It takes care for starting and stoping of the underlying `Runner` from `jest-editor-support`. It also handles `restarting` if the underlying process exits unintentionally (i.e. without being explicitly stopped) and filters out redundant events for `debuggerProcessExit` (only first event is handled, the remaining are ignored).

`JestProcessManager` manages multiple instances of `JestProcess`. In our current setup we will always have one process running at a time, but `JestProcessManager` will handle multiple process as well. `JestProcessManager` also takes care for the *watch* mode when `runAllTestsFirst` plugin setting is `true`. In such a case, it will first start a `JestProcess` to run all the tests, and when it finishes, it will automatically create another `JestProcess` in the watch mode.

Both abstractions have accompanying tests which should also document the actual usage.

This pull-request, also adds possibility to restart Jest when snapshots get updated. This is controlled by the plugin setting: `restartJestOnSnapshotUpdate`.


src/JestExt.ts
```diff
@@ -59,121 +61,90 @@ export class JestExt {
...
+            })
+            .onJestEditorSupportEvent('nonTerminalError', (error: string) => {
+              this.channel.appendLine(`Received an error from Jest Runner: ${error.toString()}`)
+            })
+            .onJestEditorSupportEvent('exception', result => {
+              this.channel.appendLine(`\nException raised: [${result.type}]: ${result.message}\n`)
+            })
+            .onJestEditorSupportEvent('terminalError', (error: string) => {
+              this.channel.appendLine('\nException raised: ' + error)
+            })
```
orta: very elegant 👍

src/JestExt.ts
```diff
@@ -210,7 +181,13 @@ export class JestExt {
         // No response == cancel
         if (response) {
           this.jestProcess.runJestWithUpdateForSnapshots(() => {
-            vscode.window.showInformationMessage('Updated Snapshots. It will show in your next test run.')
+            if (this.pluginSettings.restartJestOnSnapshotUpdate) {
+              this.jestProcessManager.stopAll()
+              this.startProcess()
+              vscode.window.showInformationMessage('Updated Snapshots and restarted Jest.')
+            } else {
+              vscode.window.showInformationMessage('Updated Snapshots. It will show in your next test run.')
```
orta: Lovely touch

src/JestExt.ts
```diff
@@ -349,14 +326,15 @@ export class JestExt {
     this.unknownItStyle = decorations.notRanItName()
   }
 
-  private shouldIgnoreOutput(text: string): boolean {
-    return text.includes('Watch Usage')
+  private shouldIgnoreOutput(_: string): boolean {
+    // this fails when snapshots change - to be revised - returning always false for now
+    // return text.includes('Watch Usage')
+    return false
```
orta: want to do anything about this?

tests/JestProcessManagement/JestProcess.test.ts
```diff
@@ -0,0 +1,283 @@
...
+        keepAlive: true,
+      })
+      jestProcess.onExit(onExit)
+      eventEmitter.emit('debuggerProcessExit')
+      eventEmitter.emit('debuggerProcessExit')
+
+      expect(onExit).toHaveBeenCalledTimes(2)
+    })
+  })
+})
```
orta: very impressive

tests/JestProcessManagement/JestProcessManager.test.ts
```diff
@@ -0,0 +1,438 @@
...
+      const jestProcess = jestProcessManager.startJestProcess({ watch: true })
+
+      eventEmitter.emit('debuggerProcessExit', jestProcess)
+
+      expect(jestProcessMock.mock.instances.length).toBe(1)
+
+      expect(jestProcessMock.mock.calls[0][0]).toHaveProperty('watchMode', true)
+    })
+  })
+})
```
orta: also very impressive

👍

src/JestExt.ts
marcinczenko: Yes. I work continuously with snapshots so I will keep eye on it. I have to look a bit closer on how output is being processed.

src/JestExt.ts
```diff
@@ -59,121 +61,90 @@ export class JestExt {
...
-    // Go!
-    if (this.pluginSettings.runAllTestsFirst) {
-      this.jestProcess.start(false)
-    } else {
-      this.startWatchMode()
-    }
+    this.jestProcessManager.startJestProcess({
+      watch: true,
+      keepAlive: true,
+      exitCallback: (_, jestProcessInWatchMode) => {
```
connectdotz: maybe I am missing something, this looks to me that all the jest output are only processed upon process exit... but during watch mode, the process will not exit isn't it? does the output get processed upon change?

src/JestProcessManagement/JestProcess.ts
```diff
@@ -0,0 +1,66 @@
...
+export class JestProcess {
+  private runner: Runner
+  private exited: boolean
+  private projectWorkspace: ProjectWorkspace
+  public keepAlive: boolean
+  public onExitCallback: Function
+  public watchMode: boolean
+
+  private startRunner() {
+    this.exited = false
```
connectdotz: you might encounter a race condition here

src/JestProcessManagement/JestProcess.ts
```diff
@@ -0,0 +1,66 @@
...
+          this.restart()
+        }
+      }
+    })
+  }
+
+  private restart() {
+    this.runner.removeAllListeners()
+    this.startRunner()
+  }
```
connectdotz: should restart() stop the running process before starting?

src/JestExt.ts
marcinczenko: This is good one. Actually it is other way around. I have assigned the handlers correctly for the watch mode (the exit handler for the first process - the one running all the tests first - provides two arguments: one for the process that just exited, second for the jest process for the watch mode that just started), but I did not assigned them for the first process that runs all the tests first.

I have just fixed it - please see the most recent commit.

src/JestProcessManagement/JestProcess.ts
marcinczenko: No, I suppose. `restart` is a private method and it is called only internally when process already exited.

src/JestProcessManagement/JestProcess.ts
marcinczenko: `startRunner` is a private method. It is only called in two circumstances:

1. When the `JestProcess` instance is created (`startRunner` is called in the constructor).
2. From method `restart`, which is also private and called only after the process already exited. Note that `restart` removes all the handlers from the underlying `Runner` before calling `startRunner`.

So, no race condition as far as I can see, but please elaborate if you thought about something else.

connectdotz: if `this.exited` is only used by the closure below, just make it a local variable, it will be captured by the closure.

src/JestProcessManagement/JestProcess.ts
connectdotz: I see, if it is only used by `startRunner()` and depends on that context, maybe better off making this method an inner function within `startRunner()`, it will prevent accidental misuse and confusion for future maintenance.

marcinczenko: I do not know... Isn't `private` not communicating that it is...private? What other people think? Could you please give an example of how you would like it to be?

src/JestProcessManagement/JestProcess.ts
marcinczenko: It is related to the other comment below. Making it part of the closure would actually indicate that we are protecting it from something at that would be misleading.

connectdotz: the code is cleaner when there is no state, or at least we should keep state as tight as possible, definitely avoid unnecessary state beyond its context. 

Suggested change:
```
private startRunner() {
    let exited = false // here
    this.runner = new Runner(this.projectWorkspace)

    this.runner.start(this.watchMode)

    this.runner.on('debuggerProcessExit', () => {
      if (!exited) { //here
        exited = true // here
        this.onExitCallback(this)
        if (this.keepAlive) {
          this.restart()
        }
      }
    })
  }
```

src/JestProcessManagement/JestProcess.ts
connectdotz: it is indeed private, but not for the context of the JestProcess class, but for the `startRunner()` method, because it only makes sense to be invoked after the process is exited as you indicated. Similar to the comment above, we need to be tight about scope/context to avoid confusion and future maintenance overhead.

Suggested change:
```
private startRunner() {
  // move the function here or inline will be fine too
  function restart() {
      this.runner.removeAllListeners()
      this.startRunner()
    }

    this.exited = false
    this.runner = new Runner(this.projectWorkspace)

    this.runner.start(this.watchMode)

    this.runner.on('debuggerProcessExit', () => {
      if (!this.exited) {
        this.exited = true
        this.onExitCallback(this)
        if (this.keepAlive) {
          restart() //here
        }
      }
    })
  }
```

connectdotz: one more question, does the keepAlive has a boundary? I have seen sometimes jest can't start, without the upper bound, will the process just keep spawning jest process until it exhausted all system resources?

marcinczenko: 1. In-line functions is not me, but I do not have time to argue about it. I will add your change (and close my eyes when typing :)).
2. Adding a boundary is a good thing. I will add that.

src/JestProcessManagement/JestProcess.ts
marcinczenko: Ok, the argument about keeping the state to the minimum is speaking to me. I will add that.

src/JestProcessManagement/JestProcess.ts
connectdotz: > In-line functions is not me

I don't like nested function either, but I dislike out-of-context/running-away-state OOP more ;-)... good news is we don't need a nested function here at all, giving it is only used in 1 place and is only 2 lines, you can further simplify it:
```
private startRunner() {
    let exited = false
    this.runner = new Runner(this.projectWorkspace)

    this.runner.start(this.watchMode)

    this.runner.on('debuggerProcessExit', () => {
      if (!exited) {
        exited = true
        this.onExitCallback(this)
        if (this.keepAlive) {
         // here, even simplier...
          this.runner.removeAllListeners()
          this.startRunner() 
        }
      }
    })
  }
```
hopefully that will keep your eyes open when you type... ;-)

marcinczenko:
Is there anything that is still blocking this to be merged?

seanpoulter:
If anyone else wants to take on the merge, feel free.
I haven't had a moment since the discussion with @connectdotz and changes from the weekend.

src/JestExt.ts
```diff
@@ -59,119 +61,95 @@ export class JestExt {
...
+    // thanks Qix, http://stackoverflow.com/questions/25245716/remove-all-ansi-colors-styles-from-strings
+    const noANSI = message.replace(/[\u001b\u009b][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g, '')
+    if (noANSI.includes('snapshot test failed')) {
+      this.detectedSnapshotErrors()
+    }
 
-        if (maxRestart-- <= 0) {
-          console.warn('jest has been restarted too many times, please check your system')
-          status.stopped('(too many restarts)')
-          return
```
seanpoulter: We're missing this behavior in the PR. When the Jest process in watch mode fails:
* the status bar spinner will keep spinning with "Starting watch mode" shown
* there's no annunciation on the output channel

src/JestExt.ts
```diff
@@ -59,119 +61,95 @@ export class JestExt {
...
+          this.channel.appendLine('Finished running all tests. Starting watch mode.')
+          status.running('Starting watch mode')
 
-  private closeJest() {
-    if (!this.jestProcess) {
-      return
-    }
-    this.forcedClose = true
-    this.jestProcess.closeProcess()
+          this.assignHandlers(this.jestProcess)
```
seanpoulter: When watch mode fails, this output is only assigned once.

src/JestExt.ts
marcinczenko: No, no really progress on it. I was a bit idle waiting for this PR to be finished and merged. It takes very long time and it is a bit demotivating so I started to put my attention somewhere else for a moment. I also wasn't sure what will be the effect of some other investigations on how to communicate with Jest Process that was mentioned in some other discussions before. For this reason I did not rush up. I am less concerned with the console output, which I found buggy anyway - it seems to swallow lots of the contents because of this "Watch Usage" filtering? I was more concerned that Snapshots work. So maybe we keep snapshots working and we open the issue on filtering the output, which I think requires review anyway.

package.json
```diff
@@ -96,6 +96,11 @@
           "description": "Whether snapshot previews should show",
           "type": "boolean",
           "default": true
+        },
+        "jest.restartJestOnSnapshotUpdate": {
+          "description": "Restart Jest runner after updating the snapshots",
+          "type": "boolean",
+          "default": false
```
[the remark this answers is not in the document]
marcinczenko: I wanted this change to be not invasive so that current users do not get something they do not expect. I am open to making it a default if there is agreement.

src/JestExt.ts
marcinczenko: This is good point. The user should be notified. I will add this behaviour.

src/JestExt.ts
marcinczenko: You might be right. Let me investigate this.

src/JestProcessManagement/JestProcess.ts
```diff
@@ -0,0 +1,62 @@
+import { Runner, ProjectWorkspace } from 'jest-editor-support'
+
+export class JestProcess {
+  static readonly keepAliveLimit = 5
+  private runner: Runner
+  private projectWorkspace: ProjectWorkspace
+  public keepAliveCounter: number
+  public onExitCallback: Function
```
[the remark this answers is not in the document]
marcinczenko: Should be possible :).

src/JestProcessManagement/JestProcess.ts
```diff
@@ -0,0 +1,62 @@
...
+  }: {
+    projectWorkspace: ProjectWorkspace
+    watchMode?: boolean
+    keepAlive?: boolean
+  }) {
+    this.watchMode = watchMode
+    this.projectWorkspace = projectWorkspace
+    this.keepAliveCounter = keepAlive ? JestProcess.keepAliveLimit : 1
+
+    this.startRunner()
```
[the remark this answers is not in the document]
marcinczenko: This one I would prefer to keep in the constructor. From my perspective it simplifies the design by removing some ugly corner cases against which I would have to perform additional checks. I like the fact that thanks to this API is very narrow and safe.

If this really feels not right for you, let me know. Then I will try to reorganise things a bit. On the other hand I am not convinced if this is right investment at the moment. There are other things I would prefer to do, and this feels really good enough. `JestProcess` will most probably undergo significant changes anyway when we find a better way to talk to jest.

src/JestProcessManagement/JestProcessManager.ts
```diff
@@ -0,0 +1,107 @@
...
+    projectWorkspace: ProjectWorkspace
+    runAllTestsFirstInWatchMode?: boolean
+  }) {
+    this.projectWorkspace = projectWorkspace
+    this.runAllTestsFirstInWatchMode = runAllTestsFirstInWatchMode
+  }
+
+  private startJestProcessInWatchMode(exitCallback, keepAlive) {
+    return this.handleNonWatchMode(true, exitCallback, keepAlive)
+  }
```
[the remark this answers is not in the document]
marcinczenko: Oh, nothing special - just the effect of refactoring and apparently I did not realised that the naming is not really to the point. I will give it a better name.

src/JestProcessManagement/JestProcessManager.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  private onJestProcessExit(jestProcess, exitCallback, keepAlive) {
+    const jestProcessInWatchMode = this.startJestProcessInWatchMode(exitCallback, keepAlive)
```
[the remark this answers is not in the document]
marcinczenko: Yes, after I refactor the naming it might be better to inline it.

src/JestExt.ts
seanpoulter: > It takes very long time and it is a bit demotivating

Yea, sorry for the delay. Life got busy, and when I did find some time it wasn't enough to get through a PR of this size. It's even slower when you aren't familiar with the new or old code, and we don't have integration tests to make sure we've preserved behavior.
<br/>

> I am less concerned with the console output, which I found buggy anyway - it seems to swallow lots of the contents because of this "Watch Usage" filtering?

That's fair. I've reverted it and opened #228 for the filtering. We filter out the Watch Usage prompt since it's extra noise we don't need to see. It shouldn't swallow anything else.

src/JestProcessManagement/JestProcess.ts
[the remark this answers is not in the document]
seanpoulter: > This one I would prefer to keep in the constructor. ... There are other things I would prefer to do, and this feels really good enough.

Sure, let's leave it for now.

seanpoulter:
Let me know if I can help with any of those comments @marcinczenko.

src/JestProcessManagement/JestProcess.ts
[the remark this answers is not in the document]
marcinczenko: Done

src/JestProcessManagement/JestProcessManager.ts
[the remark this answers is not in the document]
marcinczenko: I refactored it a bit. Do you think it is better now?

src/JestExt.ts
marcinczenko: You are right @seanpoulter. So I want to avoid adding more callbacks and taken that we already have separate `JestProcess` to deal with this kind of things, I will record registered callbacks and restore them when process is restarted inside `JestProcess` class. Feels like a right place to handle that. I will have it tomorrow.

src/JestExt.ts
marcinczenko: Working on it - when the restart threshold has been reached the callback will be called so that we can update the channel and the status bar appropriately.

src/JestProcessManagement/JestProcessManager.ts
```diff
@@ -82,9 +74,21 @@ export class JestProcessManager {
...
+        })
+        exitCallback(exitedJestProcess, jestProcessInWatchMode)
+      })
     } else {
-      return this.handleNonWatchMode(watch, exitCallback, keepAlive)
+      return this.run({
+        watch: watch,
+        keepAlive: keepAlive,
+        exitCallback: exitCallback,
+      })
```
seanpoulter: Thanks @marcinczenko. That's a lot easier to read.

marcinczenko: @seanpoulter Sorry for small delay on my side this time. The remaining changes are on the way!

seanpoulter: Don't be sorry @marcinczenko. It's a big change! After I wrap up some integration tests for `jest-editor-support`, I'll try and write some integration tests for the status bar and output channel.

marcinczenko: Just added restoring jest editor support events. These are not big changes anymore, bit to write tests and get it tested always requires a bit more time. The last thing to add is to notify the user that the allowed number of Jest process restarts has been reached. I hope to have it done by the end of Monday. Then I think we will be good to go ;).

src/JestExt.ts
marcinczenko: Ok, so I suggest the following. For processes in watch mode I call exit callback only if they failed for a number of times (`JestProcesses.keepAliveLimit`). In the response to this, I update the channel with text *Starting Jest in Watch mode failed too many times and has been stopped. Please check your system configuration.* and I set the status bar to `status.stopped()`. This way the user will see that jest is no longer running and should start jest again. We also do not have additional callback and just use one. I will still rebase everything (if needed), and give it a spin locally today to see if nothing unexpected happens.

marcinczenko: Hmm...there is still one issue with this. When stopping Jest explicitly, we will get a misleading message in the channel. I suppose I will need to add an argument to the stop handler to mark that it is the result of explicit user action.

marcinczenko: ok - there is a better way without adding any arguments. Will fix that in a moment.

marcinczenko: Ok, maybe there is a better way to accomplish that without adding any extra argument to the handlers. Let me check.

marcinczenko: Ok, done.

marcinczenko: One more small thing. Just to see how it feels, on a separate branch I created a version where `JestProcess.stop` returns a promise. I tested it out when updating snapshots and it feels quite nice. I remember that @connectdotz proposed more synchronous flow. If you think it is better like this, I will probably add a test for this. Let me know. Here is the link to the commit on a separate branch: https://github.com/marcinczenko/vscode-jest/commit/5727128a8e0c3557eb40b2024d138ec6801802a8

marcinczenko: @seanpoulter I am running it now for two days using the version where `JestProcess.stop` returns a promise and it feels quite good. Never had any surprise. I will merge the "promise" version to the PR and I hope we will be able to merge it quickly (I would like to avoid another rebasing...). I could then proceed to issue #215 as this effectively prevents snapshots from working (I am testing snapshots by modifying the `shouldIgnoreOutput()` function).

seanpoulter: Sounds good. I'll wrap up my own dev efforts then focus on the review. I'm pretty short on spare time these days, so sorry in advance for it taking a while. If anyone else has the bandwidth, feel free to take the lead.

src/JestProcessManagement/JestProcess.ts
```diff
@@ -0,0 +1,80 @@
...
+  private jestSupportEvents: Map<string, (...args: any[]) => void>
+  private keepAliveCounter: number
+  public keepAlive: boolean
+  public watchMode: boolean
+  public stopRequested: boolean
+
+  private startRunner() {
+    this.stopRequested = false
+    let exited = false
+    this.runner = new Runner(this.projectWorkspace)
```
stephtr: It would be useful to create the Runner with the shell option enabled (requires jest-editor-support 22.1.3) at least on Windows in order to fix #98.

marcinczenko: Hi @stephtr. Thanks for pointing this out. I will take a look. Hopefully, we all got some slack very soon to start review.

orta:
@seanpoulter think you can get time to review this? Otherwise I'll take a serious look at it tomorrow 👍

seanpoulter:
You might get to it first @orta. Just wrapping up my first sprint at a new job and have been spending my spare time getting up to speed with it.

seanpoulter:
Hurrah! I'll second that, thanks @marcinczenko for the thorough PR. 🎉

marcinczenko:
Wow! I am very glad it went through 🍿. Big thanks to @connectdotz, @orta, and @seanpoulter for taking your time reviewing it and giving feedback.

Now I can start dealing with other issues - especially those related to snapshots. 

We have a bit busy time at work at the moment, but very soon I should be able to do some work.

orta:
I've invited you to the org because this is a great big contribution, you're welcome to contribute as little or as much as you like - we'll be here to support you 👍

PR #218, opened 2018-01-05
Fall back to getting version directly from Jest

uucue2:
If reading the version from jest's package.json fails, try getting it directly from jest. This should fix the jest-requires-20+ message from incorrectly popping up when using the "jest.pathToJest" option.


orta:
I like this idea, any chance you can also take into account if it's done via yarn?

```sh
~/dev/projects/artsy/js/sites/metaphysics dl_2 8s
❯ yarn jest --version
yarn run v1.3.2
$ /Users/orta/dev/projects/artsy/js/sites/metaphysics/node_modules/.bin/jest --version
v21.2.1
✨  Done in 0.65s.
```

an `npm jest` should work fine though 👍 

```sh
~/dev/projects/artsy/js/sites/metaphysics dl_2
❯ npm run jest --version
5.3.0
```

seanpoulter:
Would it be easier to parse if we used `jest --showConfig`?:
```json
{
  "version": "22.0.4",
}
```

orta:
Pretty sure `--showconfig` is how we do it now

uucue2:
Well, i was thinking about it last night... would it make more sense to just use the pathToJest setting to get the right path in `helpers.ts` > `pathToJestPackageJSON` (right now it hardcodes 'node_modules/jest/package.json')? Then we can get rid of the new function altogether and only depend on getting the version from jest's package.json

uucue2:
Also, @orta , not exactly. Right now we get the path to Jest's package.json, read it in, parse the json, and pull the "version" from it:

```
const contents = readFileSync(packageJSON, 'utf8')
const packageMetadata = JSON.parse(contents)

if (packageMetadata['version']) {
  return parseInt(packageMetadata['version'])
}
```

So @seanpoulter solution would be a little cleaner since we can just use pathToJest with --showConfig

orta:
Ah, hah - so I moved [from doing the config](https://github.com/jest-community/vscode-jest/pull/15/files#diff-5782c6816a50527dc905f0367647c7c0R34) dance to reading the package.json for stability. The config isn't/wasn't meant to be publicly consumed so the keys/format changed between versions - I figured the jest package version wouldn't.

seanpoulter:
> @uucue2: would it make more sense to just use the pathToJest setting to get the right path in helpers.ts > pathToJestPackageJSON (right now it hardcodes 'node_modules/jest/package.json')? 

There was a good reminder in #98 that `pathToJest` may end up with more than just a path. For example, you may need to run Jest from a Bash shell in Windows 10. It does make more sense to me to try using `pathToJest` than the hard-coded path.
<br/>

> @orta: The config isn't/wasn't meant to be publicly consumed so the keys/format changed between versions

Sounds like we could get some practice writing integration tests! :wink:

uucue2:
@seanpoulter @orta 

I went ahead and switched over to fixing the original function that reads the version from Jest's package.json. I feel like getting the version straight from the Jest command is probably more robust (e.g. what if someone is using jspm rather than yarn or npm - it would be weird to use jspm to grab Jest, but it's a possibility), but extracting it from the package.json is faster and probably sufficient. 

If you want to switch to only using the command, I can certainly do that. I don't think it's necessary to have both methods and didn't want to overstep my bounds.

connectdotz:
If I remember this correctly the reason we were seeing lots of warning regarding jest version was because `Settings` (in `jest-editor-support`) was not able to parse the `--showConfig` output when there are noice in stdin/stdout... I vaguely remembered that even just a simple check in `Settings` to see if the output is a json can eliminate a lot of these hiccups... 

The other reason we should consider fixing the `Settings` and not just getting the version info from other place is that we do use the jest's `testMatch` pattern from the `Settings` to determine if a file is a test file ([here](https://github.com/jest-community/vscode-jest/blob/ca3e5774749553b1df21da1b400b634c6eb20baf/src/JestExt.ts#L319)). The reason we didn't hear people complain about the testMatch not-found is because `Settings` has a reasonable [default](https://github.com/facebook/jest/blob/19376c1e1e8a77717fbb7a216df92087eddbd277/packages/jest-editor-support/src/Settings.js#L54), nevertheless, it is not the actual jest config used.

> I feel like getting the version straight from the Jest command is probably more robust 

I concur. Trying to find  jest module ourselves is via `jest.pathToJest` and walking through node_modules chain could be actually a pretty hard problem to solve... besides what @seanpoulter has pointed out regarding `jest.pathToJest`, different package manager might store modules in different locations, for example, a module might be hoisted to the parent node_modules in a monorepo env... After working with react-native/webpack/monorepo/yarn projects, I came to appreciate the complexity of the module resolution, therefore I agree that it is best to avoid if possible.

My gut feeling of this issue is to fix the Settings so we can reliably get the jest config (version, testMatch etc) from `--showConfig`. What do you guys think?

src/helpers.ts
```diff
@@ -54,15 +54,24 @@ export function pathToConfig(pluginSettings: IPluginSettings) {
 }
 
 export function pathToJestPackageJSON(pluginSettings: IPluginSettings): string | null {
-  const defaultPath = normalize('node_modules/jest/package.json')
-  const craPath = normalize('node_modules/react-scripts/node_modules/jest/package.json')
+  let pathToNodeModules = join(pluginSettings.rootPath, 'node_modules')
 
+  if (pluginSettings.pathToJest) {
+    const relativeJestCmd = pluginSettings.pathToJest.split(' ')[0]
```
seanpoulter: Do we need to handle the case when the first arg is quoted? :sweat:

seanpoulter:
> @uucue2: I went ahead and switched over to fixing the original function that reads the version from Jest's package.json.

That looks pretty good @uucue2. Want me to merge that in now before we get into parsing the config?
<br/>

> @connectdotz: My gut feeling of this issue is to fix the Settings so we can reliably get the jest config (version, testMatch etc) from --showConfig. What do you guys think?

Yea! I'm keen. It's a good opportunity to close some issues and learn how to set up travis-ci to run a few tests against more than one version of Jest. You two have any idea how to do that? :nerd_face:

src/helpers.ts
uucue2: yep 😩

seanpoulter: Well, hey ... the bright side is we can use the same test case to fix [`Process.js`](https://codecov.io/gh/facebook/jest/src/master/packages/jest-editor-support/src/Process.js#L28) in `jest-editor-support`. :fire:

uucue2:
@seanpoulter  - quoted commands should now be handled properly. Yeah, i think it's a good idea to go ahead and merge this in now, then we can start dealing with the larger, underlying issue separately

PR #222, opened 2018-01-13
hierarchical module resolution when looking for jest/package.json

ronp001:
Add support for setups in which jest is installed in an ancestor directory.

(As per [node](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders) docs, the jest module can be located in a `node_modules` directory anywhere upwards in the directory hierarchy).

Note that this also relates to #218 (as inability to resolve jest version might be due to such a hierarchical setup).

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -59,6 +59,7 @@ export function pathToJestPackageJSON(pluginSettings: IPluginSettings): string |
   let updir = ''
 
   const paths = [defaultPath, craPath]
+  let max_levels = 20
   do {
     for (const i in paths) {
       const absolutePath = join(pluginSettings.rootPath, updir, paths[i])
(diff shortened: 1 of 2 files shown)


src/helpers.ts
```diff
@@ -56,13 +56,18 @@ export function pathToConfig(pluginSettings: IPluginSettings) {
...
+  do {
+    for (const i in paths) {
+      const absolutePath = join(pluginSettings.rootPath, updir, paths[i])
+      if (existsSync(absolutePath)) {
+        return absolutePath
+      }
     }
-  }
+    updir += '../'
+  } while (join(pluginSettings.rootPath, updir) !== join(pluginSettings.rootPath, updir, '../'))
```
seanpoulter: Unbounded loops make me nervous. Are there any cases where a shortcut / link could make this not end up at the root?

ronp001: Good point.  Added a failsafe mechanism.

ronp001:
I'm closing this PR because rhe code included here is incompatible with recent changes to master (which seem to have completely changed the logic which the plugin uses to look for package.json).

seanpoulter:
I think they're complimentary changes @ronp001. The recent change shuffled things around but doesn't address the problem you're trying to solve. Let's get those changes switched around and ready to go. Want me to send you a PR?

ronp001:
That will be great - thanks @seanpoulter.

seanpoulter:
I had a look at this one over the weekend @ronp001. Does this fix a bug with the extension right now that using the `rootPath` or `pathToJest` settings doesn't cover? It seems to me like we're improving the "default" behavior a bit.

PR #243, opened 2018-01-31
add jest-cli to pathToJestPackgeJSON #trivial

TreTuna:
VSC was giving an error when a user had "jest-cli" installed instead of simple "jest" package. This adds support for that circumstance.


orta:
This looks 👍 to me

PR #252, opened 2018-02-05
Strips testNames so they can be used as regex closes #251

### LINKED ISSUE #251: Debug Requires test names to be a valid regex
I have a test named `reset() -> Reset`. This really confuses the jest plugin because it takes the name of my test as a string and just assumes it's a valid regex. 

If I change my test name to:`test("reset( -> Reset", () => {})`

Debug will crash with this error:
image

BLamy:


orta:
Lovely, :

orta:
👍 thanks!

PR #268, opened 2018-02-21
Further improved detection logic for projects created by `create-react-app`

stephtr:
#266 obviously fails to detect custom forks of `react-scripts`. By comparing `scripts.test` within `package.json` instead of the dependencies, one can also automatically detect most drop-in replacements for `react-scripts` since most of them don't change the binary name (in order to be drop-in). Additionally I added detection for `react-app-rewired`.
Sorry for not thinking about that in the first place.
#trivial


orta:
No need to be sorry, this sounds good to me 👍

orta:
Shipped as 2.6.3

Issue #193, opened 2017-12-13
Debug CodeLens doesn't use jest.pathToJest


seanpoulter:
### Steps to Reproduce
- Clone `git clone https://github.com/seanpoulter/vscode-jest--193`
  - This is a bare repo bootstrapped for issue #190 with "jest.pathToJest" set and a failing test
- Open the repo in Visual Studio Code
- Click on the Debug CodeLens

### Expected Behavior
- Debug the test
- [`JestExt#resolvePathToJestBin()`](https://github.com/orta/vscode-jest/blob/e166c3718307efc85589a656dcdcd2fd8808c175/src/JestExt.ts#L491) can handle more complicated commands like `npm test --`

### Actual Behavior
![image](https://user-images.githubusercontent.com/2585460/33959961-26b2cecc-e017-11e7-9271-b3beecdc1317.png)

stephtr:
This is now by design (#287).

Issue #233, opened 2018-01-28
Feature request: Add a command to open the file under test or test for file


seanpoulter:
### Expected Behavior

I'd love to have a command opens the related test or file for the current document. This should work both ways:
- When editing source code, open the test
- When editing a test, open the source code
Since we won't be certain, the command could open up a quick pick menu showing the imported files.

I'm not sure if that falls under the scope of this extension, but I find myself wishing this was a thing quite often so I'm pretty motivated to add it.

orta:
Yeah, makes sense to me to be in here. This code exists somewhere in jest, note that there can be multiple test files for one file, so ideally looping through them>

seanpoulter:
Good point @orta. I'll see if there's something I can pull in from Jest.

PR #271, opened 2018-02-22
Improve debugging of tests

### LINKED ISSUE #239: Debug w/ Create React App does not seem to work.
### Environment

1. `node -v`: 8.9.4
2. `npm -v`: 5.6.0
4. `npm ls react-scripts` (if you haven’t ejected): react-scripts@1.0.17

1. Operating system: OSX 10.13.2

### Steps to Reproduce

1. open vscode and ensure I have the latest `vscode-jest` installed
2. `npx create-react-app my-app`
3. `cd my-app && yarn add @types/jest`
4. Create the `.vscode/launch.json` file and add the following configuration, from the `create-react-app` readme. 

```
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Debug CRA Tests",
      "type": "node",
      "request": "launch",
      "runtimeExecutable": "${workspaceRoot}/node_modules/.bin/react-scripts",      
      "args": [
        "test",
        "--runInBand",
        "--no-cache",
        "--env=jsdom"
      ],
      "cwd": "${workspaceRoot}",
      "protocol": "inspector",
      "console": "integratedTerminal",
      "internalConsoleOptions": "neverOpen"
    }
  ]
}
```
source: https://github.com/facebook/create-react-app/blob/ed5c48c81b2139b4414810e1efe917e04c96ee8d/packages/react-scripts/template/README.md#debugging-tests-in-visual-studio-code

5. Add the following test to `src/App.test.js`

```
it('SIMPLE TEST', () => {
  let x = 1;
  let y = 2;
  expect(1).toEqual(2);
});
```
6. Add a breakpoint, and Click the `Debug` lens above the test. 

### Expected Behavior

The debug overlay to open and the debugger to break.

### Actual Behavior

`Error: Cannot find jest.js file`

--- 

Note: Following some research elsewhere on this issue queue and elsewhere I've also tried setting my `jest.pathToJest` in my `.vscode/settings.json` to a number of different settings with no luck. 

Reference: 

* https://github.com/jest-community/vscode-jest/issues/113
* https://github.com/jest-community/vscode-jest/issues/125

* https://github.com/facebook/create-react-app/issues/2709
* https://github.com/facebook/create-react-app/issues/594

* https://stackoverflow.com/questions/47575916/vscode-says-it-cannot-find-jest-js-file

stephtr:
Closed in favor of #287 
---
Previously debugging of tests was only possible when `pathToJest` referenced to the jest binary.
This PR should enable support for other JS files or `npm test --`, including projects which have been generated using `create-react-app`. (fixes #239 and in most cases #193)

Since this change misses quite for sure some corner cases, I would be happy for some pairs of eyes looking over it or trying it out, even though this change shouldn't degrade support.111

### DIFF
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -445,7 +445,7 @@ export class JestExt {
   /**
    * Primitive way to resolve path to jest.js
    */
-  private resolvePathToJestBin() {
+  private resolveTestProgram(): { program: string; args: string[]; isCreateReactApp: boolean } {
     // Let's start with the given command
     let jest = this.workspace.pathToJest
     let basename = path.basename(jest)
@@ -569,7 +569,7 @@ export class JestExt {
   public runTest = (fileName: string, identifier: string) => {
     const restart = this.jestProcessManager.numberOfProcesses > 0
     this.jestProcessManager.stopAll()
-    const launcher = this.resolvePathToJestBin()
+    const launcher = this.resolveTestProgram()
     if (!launcher) {
       console.log("Could not find Jest's CLI path")
       return
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -586,6 +586,7 @@ export class JestExt {
 
     const port = Math.floor(Math.random() * 20000) + 10000
     let runtimeArgs = ['--inspect-brk=' + port]
+    const env: any = {}
 
     if (launcher.isCreateReactApp) {
       // If the project has been created by create-react-app,
@@ -598,6 +599,9 @@ export class JestExt {
       // arguments for generating the new node process.
       args = runtimeArgs.concat(args)
       runtimeArgs = []
+      // `react-scripts` automatically appends the `--watch` flag
+      // if CI isn't set.
+      env.CI = 'VSCode-test-debugger'
     }
 
     const configuration = {
@@ -612,7 +616,7 @@ export class JestExt {
       console: 'integratedTerminal',
       smartStep: true,
       sourceMaps: true,
-      env: { CI: 'VSCode-test-debugger' },
+      env,
     }
 
     const handle = vscode.debug.onDidTerminateDebugSession(_ => {
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -474,72 +474,80 @@ export class JestExt {
     const extension = path.extname(basename)
     jest = path.join(path.dirname(jest), basename)
 
-    switch (extension) {
-      case 'js': {
-        return jest
-      }
+    let runCommand = ''
 
-      case 'cmd': {
-        /* i need to extract '..\jest-cli\bin\jest.js' from line 2
-
-        @IF EXIST "%~dp0\node.exe" (
-          "%~dp0\node.exe"  "%~dp0\..\jest-cli\bin\jest.js" %*
-        ) ELSE (
-          @SETLOCAL
-          @SET PATHEXT=%PATHEXT:;.JS;=;%
-          node  "%~dp0\..\jest-cli\bin\jest.js" %*
-        )
-        */
-        const line = fs.readFileSync(jest, 'utf8').split('\n')[1]
-        const match = /^\s*"[^"]+"\s+"%~dp0\\([^"]+)"/.exec(line)
-        return path.join(path.dirname(jest), match[1])
-      }
-
-      case '': {
-        /* file without extension uses first line as file type
-           in case of node script i can use this file directly,
-           in case of linux shell script i need to extract path from line 9
-        #!/bin/sh
-        basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
-
-        case `uname` in
-            *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
-        esac
-
-        if [ -x "$basedir/node" ]; then
-          "$basedir/node"  "$basedir/../jest-cli/bin/jest.js" "$@"
-          ret=$?
-        else
-          node  "$basedir/../jest-cli/bin/jest.js" "$@"
-          ret=$?
-        fi
-        exit $ret
-        */
-        const lines = fs.readFileSync(jest, 'utf8').split('\n')
-        switch (lines[0]) {
-          case '#!/usr/bin/env node': {
-            return jest
-          }
-
-          case '#!/bin/sh': {
-            const line = lines[8]
-            const match = /^\s*"[^"]+"\s+"\$basedir\/([^"]+)"/.exec(line)
-            if (match) {
-              return path.join(path.dirname(jest), match[1])
-            }
-
-            break
-          }
+    try {
+      switch (extension) {
+        case 'js': {
+          runCommand = jest
+          break
         }
 
-        break
-      }
+        case 'cmd': {
+          /* i need to extract '..\jest-cli\bin\jest.js' from line 2
+  
+          @IF EXIST "%~dp0\node.exe" (
+            "%~dp0\node.exe"  "%~dp0\..\jest-cli\bin\jest.js" %*
+          ) ELSE (
+            @SETLOCAL
+            @SET PATHEXT=%PATHEXT:;.JS;=;%
+            node  "%~dp0\..\jest-cli\bin\jest.js" %*
+          )
+          */
+          const line = fs.readFileSync(jest, 'utf8').split('\n')[1]
+          const match = /^\s*"[^"]+"\s+"%~dp0\\([^"]+)"/.exec(line)
+          runCommand = path.join(path.dirname(jest), match[1])
+          break
+        }
 
-      case 'npm test --':
-      case 'npm.cmd test --': {
-        vscode.window.showErrorMessage('Debugging of tasks is currently only available when directly running jest!')
-        return undefined
+        case '': {
+          /* file without extension uses first line as file type
+             in case of node script i can use this file directly,
+             in case of linux shell script i need to extract path from line 9
+          #!/bin/sh
+          basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
+  
+          case `uname` in
+              *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
+          esac
+  
+          if [ -x "$basedir/node" ]; then
+            "$basedir/node"  "$basedir/../jest-cli/bin/jest.js" "$@"
+            ret=$?
+          else
+            node  "$basedir/../jest-cli/bin/jest.js" "$@"
+            ret=$?
+          fi
+          exit $ret
+          */
+          const lines = fs.readFileSync(jest, 'utf8').split('\n')
+          switch (lines[0]) {
+            case '#!/usr/bin/env node': {
+              runCommand = jest
+              break
+            }
+
+            case '#!/bin/sh': {
+              const line = lines[8]
+              const match = /^\s*"[^"]+"\s+"\$basedir\/([^"]+)"/.exec(line)
+              if (match) {
+                runCommand = path.join(path.dirname(jest), match[1])
+              }
+
+              break
+            }
+          }
+
+          break
+        }
       }
+    } catch {
+      vscode.window.showErrorMessage('Cannot debug jest command!')
+      return undefined
+    }
+
+    if (runCommand !== '') {
+      return { program: runCommand, args: args }
     }
 
     vscode.window.showErrorMessage('Cannot find jest.js file!')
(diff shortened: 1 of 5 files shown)


orta:
OK, I can't quite grok this in my head, will try find time over the weekend and test it locally

src/JestExt.ts
```diff
@@ -445,79 +445,121 @@ export class JestExt {
...
     }
 
-    const basename = path.basename(jest)
-    switch (basename) {
-      case 'jest.js': {
-        return jest
-      }
+    // We hope for the best that the filename doesn't contain any spaces and separate
+    // the basename from the arguments.
+    const args = basename.split(' ')
```
stephtr: Adding an external dependency would probably be the best way for a correct separation of arguments (including quotes)?

stephtr: (For example https://www.npmjs.com/package/string-argv)

seanpoulter: The root cause of this problem is `jest.pathToJest` is ambiguous as a string, and hard to parse into the parameters we need to provide to `child_process.spawn`. Why not remove the problem by asking for the type signature we need: `{ cmd: String, args: String[] }`.

You're going to see this again soon as a suggestion to fix the regression we've introduced on Windows.

stephtr: That would work for the settings, but not, if one sets it to npm test, in which case we would still end with the command string from package.json.

Which regression do you mean?

seanpoulter: I'm not sure I understand the issue. Is the problem that a user who has requested that we run the "npm test" script to start Jest is forcing us to find the path to the JavaScript file?

Would the args be passed to `node` if we spawned that command?

connectdotz:
Hi, sorry I was late for this thread and many others earlier... 

Looks like recently there are a few CRA related issues (#244, #239), it might be a good time to take a step back and discuss if keep adding extra patterns and logic to handle this ever-widening landscape is the only way we should consider... 

JavaScript/Typescripts environments will always have lots of variations, platforms, libraries, and that is what made it great. I think we can all agree that **vscode-jest can't resolve every possible variation people can run or debug jest tests**. We need to draw a clear boundary for what vscode-jest can do, and provide flexible hooks to allow users adapt beyond. Otherwise, this tool might be risking in growing complexity, reducing stability and still falling short of addressing users' needs...

in #244, I think @seanpoulter [suggestion](https://github.com/jest-community/vscode-jest/issues/244#issuecomment-362275787) to use `jest.pathToJest` is perfectly fine, actually, I like it better than trying to locate jest binary ourselves, which will most likely fail in monorepo projects, symlinks and possibly more (yarn vs npm?)... Instead of elaborated "guesses" that might not be right, IMHO, we should focus on helping people to use (or improve) the existing hooks like `jest.pathToJest` to address their specific needs/env; after all, users know their environments much better than we can ever do.

Furthermore, in this [comment](https://github.com/jest-community/vscode-jest/issues/244#issuecomment-363067768), @rrousselGit already has a working debug config. Similar to the argument above, why guess when the user already had the answer? Instead of keep fixing our internally generated debug config, I think we should focus on launching the debug session with the ability to take user provided debug config. This issue is even more obvious in #233 where @justinlevi used the debug config suggested in react-scripts [doc](https://github.com/facebook/create-react-app/blob/ed5c48c81b2139b4414810e1efe917e04c96ee8d/packages/react-scripts/template/README.md#debugging-tests-in-visual-studio-code).

In that spirit, I think a more efficient "catch-all" solution is not to rely on `resolvePathToJestBin` nor a special logic to detect CRA or any other custom environments; but let users plug in their custom settings via vscode-jest hooks, like what @orta has done with `jest.pathToJest`.

can it be done for debug? I hope so... vscode recommended debug plugin implementation via [DebugConfigurationProvider](https://code.visualstudio.com/docs/extensions/example-debuggers#_using-a-debugconfigurationprovider) interface that can intercept and inject/modify debug config. The official [example](https://github.com/Microsoft/vscode-mock-debug/blob/042d19a27a8e3a08f27a24110506b53fbecc75ce/src/extension.ts#L34) seems pretty straightforward and can satisfy our requirements: 
1. work with user-supplied debug configs.
1. inject a debug config if none provided.

A little bit more thoughts on possible implementation: 

- **provide a default config if none is provided by user**
  similar to what we already have, but instead of using `resolvePathToJestBin()` to resolve "program", we could do this in `DebugConfigurationProvider.resolveDebugConfiguration`:
  ```
  program = pathToJest.debug || pathToJest
  ```
  `pathToJest.debug` (placeholder, a new vscode-jest hook) allows user to provide debug specific jest program, if different from `pathToJest`.

- **debug with custom debug config**
  we could introduce a new hook (say `debugConfigName` for now), and start the debug session with it if specified. Inside the `DebugConfigurationProvider.resolveDebugConfiguration`, all we have to do is to inject the dynamic params like `testNamePattern` into the `config.args`, relatively simple...

- **documentation**
  clearly call out vscode-jest philosophy, boundary, defaults and how people can help themselves beyond the default settings, such as CRA. We can provide debug config examples and tap into the wonderful community knowledge already available, such as [react-scripts](https://github.com/facebook/create-react-app/blob/ed5c48c81b2139b4414810e1efe917e04c96ee8d/packages/react-scripts/template/README.md#debugging-tests-in-visual-studio-code), [vscode](https://github.com/Microsoft/vscode-recipes/tree/master/debugging-jest-tests)...

Sorry for a rather long comment, just to provide an alternative way to reason about these issues/solutions...

stephtr:
> Looks like recently there are a few CRA related issues (#244, #239), it might be a good time to take a step back and discuss if keep adding extra patterns and logic to handle this ever-widening landscape is the only way we should consider…

I'd be glad to.

> […] vscode-jest can't resolve every possible variation people can run or debug jest tests. We need to draw a clear boundary for what vscode-jest can do […]

In my opinion that boundary would be to include support for the most common project generators. People mainly use CRA because of the unnecessity of dealing with configurations, in contrast to those using monorepos.

> This issue is even more obvious in #233 where @justinlevi used the debug config suggested in react-scripts doc.

Ah, in the meantime the configuration has changed, so even for CRA apps it would be enough to just run `npm test` with the necessary test parameters appended. For me that is an additional argument for **defaulting `pathToJest` to `npm test --`**, which then should **work for most projects out-of-box**, without any specific code for CRA in this repo.

> pathToJest.debug (placeholder, a new vscode-jest hook) allows user to provide debug specific jest program, if different from pathToJest.

In which case would `pathToJest.debug` be different from `pathToJest`? For example when using CRA, one could set `pathToJest` to `npm test --`, but then he have to set `pathToJest.debug` to `react-scripts test`, in which case he could have nevertheless set both values to `react-scripts test` the first time.

> debug with custom debug config

That would be an additional option, but independently I still would like to achieve that this extension works for 90% of the most common setups without any configuration, if that would be possible by using the test script from package.json.

> documentation

In my opinion (that was also my first impression when using this extension) documentation should be definitely expanded.

----

One alternative way I could think of is to ask the user a few questions when opening a project containing jest for the first time, similar to trying to debug a new project for the first time, like:
'We detected that this project utilizes jest. In order to monitor and debug tests, we need you to answer a few questions for configuration [Configure]'
'Which command has to be executed for testing? ' - show a list of suggestions based on the project's content (CRA, npm test script, jest, ...)
'Do you want to automatically collect code coverage? Enabling it could result in higher CPU usage on larger projects. [yes] [no]'
…
The answers would be saved to the settings file; in contrast to directly modifying settings we would have the opportunity to present the user a few educated guesses, enhancing the out-of-box experience for the user.

----

Independent of the decision I would like to add a setting `automatically collect code coverage`, which would just add `--collectCoverage` as an argument. When the user then toggles coverage overlay while that setting isn't set, we could ask, if we should switch on the setting for him.

connectdotz:
> In my opinion that boundary would be to include support for the most common project generators. 

To keep up with external library/platform/generator is not trivial. I still remember when jest 21.0.0 moved to the new config format, all vscode-jest users with new jest were broken and blocked until we can release the patch.  Imagine after we commit this PR and CRA changed the config, now do we need to add more logic to detect CRA version in order to pass the right params? It's a slippery slope... dependency is expensive, lesser the better. Jest is a necessary dependency, but nothing else is. While we all want to be as user-friendly as possible, we need to be conscious of the price we will be paying.

> Ah, in the meantime the configuration has changed, so even for CRA apps it would be enough to just run npm test with the necessary test parameters appended.

Does this mean we can rollback #266 and `testCommandIsCreateReactApp`?

> For me that is an additional argument for defaulting pathToJest to npm test --, which then should work for most projects out-of-box, without any specific code for CRA in this repo.

perfect, I am perfectly fine with whatever default value we decide for pathToJest. Does this mean that we can also get rid of `resolveTestProgram`? 

> In which case would pathToJest.debug be different from pathToJest?

if pathToJest defaults to `npm test --` as you suggested, and the debug config "program" wants to use plain jest instead, one can set `jest.pathToJest.debug` to `node_modues/.bin/jest.js`, for example. It is an optional setting, could default to undefined if appropriate. It is mainly to provide flexibility to let **users** override pathToJest during debug session if needed. 
 
> when using CRA, one could set pathToJest to npm test --, but then he have to set pathToJest.debug to react-scripts test, in which case he could have nevertheless set both values to react-scripts test the first time.

Not sure what is the problem here? The user doesn't have to set `pathToJest.debug` if it is the same as 'pathToJest'. `program = pathToJest.debug || pathToJest`

> debug with custom debug config... That would be an additional option, but independently I still would like to achieve that this extension works for 90% of the most common setups without any configuration, if that would be possible by using the test script from package.json.

custom debug config, like pathToJest.debug, is optional: `debug config = custom-debug-config  || generated-debug-config`. I completely agree that most use cases shouldn't need to provide its own debug config. With reasonable pathToJest and pathToJest.debug, the internally generated one should work just fine for most use cases. Do you think there are some other things we should do in order to support the 90% use cases?

> One alternative way I could think of is to ask the user a few questions when opening a project containing jest for the first time...

This might be a reasonable alternative. However, users might change their packages after the initial opening, so would need to update the config accordingly. Thus you might want to make it a standalone executable, like `npm init`, which can be invoked on demand.

rrousselGit:
Is this PR a solution to the `Debugging of tasks is currently only available when directly running jest!` ?

stephtr:
> Imagine after we commit this PR and CRA changed the config, now do we need to add more logic to detect CRA version in order to pass the right params?

One of the ideas of CRA is that they hide all of the new features and configuration changes behind their `react-scripts` package. But it's true, if we have to separately handle CRA apps, there is always the possibility that one day functionality breaks.

> Does this mean we can rollback #266 and testCommandIsCreateReactApp?

Concerning running the tests (in background) `testCommandIsCreateReactApp` is only necessary for switching the default value of `jest.pathToJest` to `npm test --`, such that running the tests is possible with CRA apps without changing the configuration. Personally I would simply change the default value of `pathToJest` to `npm test --`, which at least in that case would eliminate the necessarity of detecting CRA apps. Other projects would then have to add `scripts: { test: 'jest' }` (if we don't include a simple fallback) or similar to their package.json (which most of them probably already have) or manually change pathToJest back to `node_modules/.bin/jest`.
#266 on its own shouldn't be rolled back since it is only an improvement of the detection of CRA. If we remove the CRA dependency, we should remove the whole detection/switching logic.

> if pathToJest defaults to npm test -- as you suggested, and the debug config "program" wants to use plain jest instead, one can set jest.pathToJest.debug

I'm fine with adding such a setting (or including it in the debug-config), I just thought that in such case `pathToJest` could also be set to plain jest.

> Do you think there are some other things we should do in order to support the 90% use cases?
I would reckon to have a default setting, which would work with plain jest and with CRA apps. However comparing the debug config for [jest only](https://github.com/Microsoft/vscode-recipes/tree/master/debugging-jest-tests#configure-launchjson-file-for-yur-test-framework) and [jest on CRA](https://github.com/facebook/create-react-app/blob/ed5c48c81b2139b4414810e1efe917e04c96ee8d/packages/react-scripts/template/README.md#debugging-tests-in-visual-studio-code) (one time `program`, one time `runtimeExecutable`, one time `node_modules/<package>/<path-tobin) I can hardly think of any such solution.

connectdotz:
Let's get back to the code... @stephtr, will you be interested in taking a shot at clean-up and moving the debug functionality into [DebugConfigurationProvider](https://code.visualstudio.com/docs/extensions/example-debuggers#_using-a-debugconfigurationprovider)? This will pave the way for further debug enhancement such as supporting user-defined debug config.

stephtr:
Wouldn't it be less work to directly include user-defined debug configurations? Did I understand it correctly, we would create a DebugConfigurationProvider which automatically appends the flags specifying the test to be run, the user creates a suitable debug configuration with a specific name and then we would run that configuration as soon as the user clicks "Debug test"?

Unfortunately I won't be available the coming days, but I could try it the week after next.

src/JestExt.ts
```diff
@@ -527,32 +569,54 @@ export class JestExt {
...
       args,
-      runtimeArgs: ['--inspect-brk=' + port],
+      runtimeArgs: runtimeArgs,
       port,
       protocol: 'inspector',
       console: 'integratedTerminal',
       smartStep: true,
       sourceMaps: true,
+      env,
     }
```
nevir: It'd be helpful to also let us override `runtimeExecutable`, for cases where we need to pass `--inspect-brk` through to a child process

(example use case: https://github.com/nevir/lib-boundless/blob/master/scripts/jest.js)

stephtr: That would be the plan, since that is also what we need for supporting create-react-app apps.

nevir: Ah, sweet

stephtr:
Sorry that it took me so long for finding time.
Today I had some time for experimenting with `DebugConfigurationProvider`, but at the moment I'm kind of stuck. Usually when one writes such a provider, he also has to supply an appropriate Debug Adapter. Since we would like to use node, my first action in `resolveDebugConfiguration(...)` was to set `debugConfiguration.type = 'node'` (from the name of my DebugConfigurationProvider back to default), however that is already too late. Is there another way for providing the default node Debug Adapter or would it be easier to move the debug configuration into VS Code's settings?
[stephtr/vscode-jest/debug-support](https://github.com/stephtr/vscode-jest/commit/f247cdb15ba10e10c40b6f1dd9ba53a003abf45d) is my current state.

PR #287, opened 2018-03-26
Outsource debug configuration to launch.json

stephtr:
As often mentioned we need a possibility for debugging tasks if the script to be run is not the standard jest binary. The discussion #271 leaned towards using the DebugConfiguration instead of the `pathToJest` setting, since one then has more control over the settings used, which simplifies the core extension when including support for CreateReactApps and similar.
With this PR debugging tests should work for most jest and CreateReactApp projects out of the box, in general one has to add a launch configuration named `vscode-jest-tests` (of type `node`).
Still missing are tests for `DebugConfigurationProvider.ts`, which I would add, if the general design of this PR is okay.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,7 +126,7 @@
     ],
     "debuggers": [
       {
-        "type": "vscode-jest-tests",
+        "type": "node",
         "label": "Debug Jest Tests using vscode-jest"
       }
     ]
--- a/package.json
+++ b/package.json
@@ -123,6 +123,12 @@
         "command": "io.orta.jest.coverage.toggle",
         "title": "Jest: Toggle Coverage Overlay"
       }
+    ],
+    "debuggers": [
+      {
+        "type": "vscode-jest-tests",
+        "label": "Debug Jest Tests using vscode-jest"
+      }
     ]
   },
   "lint-staged": {
--- a/src/DebugConfigurationProvider.ts
+++ b/src/DebugConfigurationProvider.ts
@@ -35,13 +35,13 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
       name: 'vscode-jest-tests',
       request: 'launch',
       args: ['--runInBand'],
-      cwd: '${workspaceRoot}',
+      cwd: '${workspaceFolder}',
       console: 'integratedTerminal',
       internalConsoleOptions: 'neverOpen',
     }
     const craCommand = tryGetCRACommand(folder.uri.fsPath).split(' ')
     if (craCommand.length > 1 || craCommand[0]) {
-      debugConfiguration.runtimeExecutable = '${workspaceRoot}/node_modules/.bin/' + craCommand.shift()
+      debugConfiguration.runtimeExecutable = '${workspaceFolder}/node_modules/.bin/' + craCommand.shift()
       debugConfiguration.args = [...craCommand, ...debugConfiguration.args]
       debugConfiguration.protocol = 'inspector'
     } else {
@@ -58,6 +58,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
     if (debugConfiguration.name !== 'vscode-jest-tests') {
       return debugConfiguration
     }
+    // necessary for running CRA test scripts in non-watch mode
     if (debugConfiguration.env) {
       debugConfiguration.env.CI = 'vscode-jest-tests'
     } else {
--- a/src/DebugConfigurationProvider.ts
+++ b/src/DebugConfigurationProvider.ts
@@ -1,0 +1,70 @@
+import * as vscode from 'vscode'
+import { join } from 'path'
+import { readFileSync } from 'fs'
+
+function tryGetCRACommand(rootPath: string): string {
+  // Known binary names of `react-scripts` forks:
+  const packageBinaryNames = ['react-scripts', 'react-native-scripts', 'react-scripts-ts', 'react-app-rewired']
+  // If possible, try to parse `package.json` and look for a known binary beeing called in `scripts.test`
+  try {
+    const packagePath = join(rootPath, 'package.json')
+    const packageJSON = JSON.parse(readFileSync(packagePath, 'utf8'))
+    if (!packageJSON || !packageJSON.scripts || !packageJSON.scripts.test) {
+      return ''
+    }
+    const testCommand = packageJSON.scripts.test as string
+    if (packageBinaryNames.some(binary => testCommand.indexOf(binary + ' test') === 0)) {
+      return testCommand
+    }
+  } catch {}
+  return ''
+}
+
+export class DebugConfigurationProvider implements vscode.DebugConfigurationProvider {
+  private fileNameToRun: string = ''
+  private testToRun: string = ''
+
+  public prepareTestRun(fileNameToRun: string, testToRun: string) {
+    this.fileNameToRun = fileNameToRun
+    this.testToRun = testToRun
+  }
+
+  provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken) {
+    const debugConfiguration: vscode.DebugConfiguration = {
+      type: 'vscode-jest-tests',
+      name: 'vscode-jest-tests',
+      request: 'launch',
+      args: ['--runInBand'],
+      cwd: '${workspaceRoot}',
+      console: 'integratedTerminal',
+      internalConsoleOptions: 'neverOpen',
+    }
+    const craCommand = tryGetCRACommand(folder.uri.fsPath).split(' ')
+    if (craCommand.length > 1 || craCommand[0]) {
+      debugConfiguration.runtimeExecutable = '${workspaceRoot}/node_modules/.bin/' + craCommand.shift()
+      debugConfiguration.args = [...craCommand, ...debugConfiguration.args]
+      debugConfiguration.protocol = 'inspector'
+    } else {
+      debugConfiguration.program = '${workspaceFolder}/node_modules/jest/bin/jest'
+    }
+    return [debugConfiguration]
+  }
+
+  resolveDebugConfiguration(
+    folder: vscode.WorkspaceFolder | undefined,
+    debugConfiguration: vscode.DebugConfiguration,
+    token?: vscode.CancellationToken
+  ) {
+    debugConfiguration.type = 'node'
+    if (this.fileNameToRun) {
+      debugConfiguration.args.push(this.fileNameToRun)
+      if (this.testToRun) {
+        debugConfiguration.args.push('--testNamePattern')
+        debugConfiguration.args.push(this.testToRun)
+      }
+      this.fileNameToRun = ''
+      this.testToRun = ''
+    }
+    return debugConfiguration
+  }
+}
--- a/src/DebugConfigurationProvider.ts
+++ b/src/DebugConfigurationProvider.ts
@@ -31,7 +31,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
 
   provideDebugConfigurations(folder: vscode.WorkspaceFolder | undefined, token?: vscode.CancellationToken) {
     const debugConfiguration: vscode.DebugConfiguration = {
-      type: 'vscode-jest-tests',
+      type: 'node',
       name: 'vscode-jest-tests',
       request: 'launch',
       args: ['--runInBand'],
@@ -55,7 +55,14 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
     debugConfiguration: vscode.DebugConfiguration,
     token?: vscode.CancellationToken
   ) {
-    debugConfiguration.type = 'node'
+    if (debugConfiguration.name !== 'vscode-jest-tests') {
+      return debugConfiguration
+    }
+    if (debugConfiguration.env) {
+      debugConfiguration.env.CI = 'vscode-jest-tests'
+    } else {
+      debugConfiguration.env = { CI: 'vscode-jest-tests' }
+    }
     if (this.fileNameToRun) {
       debugConfiguration.args.push(this.fileNameToRun)
       if (this.testToRun) {
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -18,6 +18,7 @@ import { readFileSync } from 'fs'
 import { CoverageMapProvider } from './Coverage'
 import { updateDiagnostics, resetDiagnostics, failedSuiteCount } from './diagnostics'
 import { DebugCodeLensProvider } from './DebugCodeLens'
+import { DebugConfigurationProvider } from './DebugConfigurationProvider'
 import { DecorationOptions } from './types'
 import { hasDocument, isOpenInMultipleEditors } from './editor'
 import { CoverageOverlay } from './Coverage/CoverageOverlay'
@@ -33,6 +34,7 @@ export class JestExt {
 
   testResultProvider: TestResultProvider
   public debugCodeLensProvider: DebugCodeLensProvider
+  debugConfigurationProvider: DebugConfigurationProvider
 
   // So you can read what's going on
   private channel: vscode.OutputChannel
@@ -69,6 +71,7 @@ export class JestExt {
 
     this.testResultProvider = new TestResultProvider()
     this.debugCodeLensProvider = new DebugCodeLensProvider(this.testResultProvider, pluginSettings.enableCodeLens)
+    this.debugConfigurationProvider = new DebugConfigurationProvider()
 
     this.jestProcessManager = new JestProcessManager({
       projectWorkspace: workspace,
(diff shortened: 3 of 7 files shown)


src/JestExt.ts
```diff
@@ -562,7 +457,13 @@ export class JestExt {
       }
     })
 
-    vscode.debug.startDebugging(vscode.workspace.workspaceFolders[0], configuration)
+    const workspaceFolder = vscode.workspace.workspaceFolders[0]
+    try {
+      await vscode.debug.startDebugging(workspaceFolder, 'vscode-jest-tests')
+    } catch {
+      const debugConfiguration = this.debugConfigurationProvider.provideDebugConfigurations(workspaceFolder)[0]
+      vscode.debug.startDebugging(workspaceFolder, debugConfiguration)
```
connectdotz: maybe we can capture error here as well and provides some helpful notes to improve the user experience...

src/DebugConfigurationProvider.ts
```diff
@@ -0,0 +1,78 @@
...
+      debugConfiguration.program = '${workspaceFolder}/node_modules/jest/bin/jest'
+    }
+    return [debugConfiguration]
+  }
+
+  resolveDebugConfiguration(
+    folder: vscode.WorkspaceFolder | undefined,
+    debugConfiguration: vscode.DebugConfiguration,
+    token?: vscode.CancellationToken
+  ) {
```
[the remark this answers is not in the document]
stephtr: Ah thanks, I just saw the warnings and thought that I maybe would use those arguments later. Is fixed.

src/JestExt.ts
stephtr: Unfortunately the exception throwing behavior of `startDebugging` isn't documented. The only case in which it throws (a `null` exception) seems to be if the requested debug configuration isn't present. If starting the debug session fails because of a wrong command in the configuration, vscode displays a more or less detailed error message by itself. The only modification I made was to show a warning message, if it falls back to the default configuration.

src/DebugConfigurationProvider.ts
```diff
@@ -0,0 +1,78 @@
...
+      name: 'vscode-jest-tests',
+      request: 'launch',
+      args: ['--runInBand'],
+      cwd: '${workspaceFolder}',
+      console: 'integratedTerminal',
+      internalConsoleOptions: 'neverOpen',
+    }
+    const craCommand = tryGetCRACommand(folder.uri.fsPath).split(' ')
+    if (craCommand.length > 1 || craCommand[0]) {
+      debugConfiguration.runtimeExecutable = '${workspaceFolder}/node_modules/.bin/' + craCommand.shift()
```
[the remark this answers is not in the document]
stephtr: On my Windows machine it works, since cmd partially supports using slashes instead of backslashes.

src/DebugConfigurationProvider.ts
```diff
@@ -0,0 +1,78 @@
...
+      debugConfiguration.args = [...craCommand, ...debugConfiguration.args]
+      debugConfiguration.protocol = 'inspector'
+    } else {
+      debugConfiguration.program = '${workspaceFolder}/node_modules/jest/bin/jest'
```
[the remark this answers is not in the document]
stephtr: 👍

package.json
```diff
@@ -123,6 +123,12 @@
         "command": "io.orta.jest.coverage.toggle",
         "title": "Jest: Toggle Coverage Overlay"
       }
+    ],
+    "debuggers": [
+      {
+        "type": "node",
+        "label": "Debug Jest Tests using vscode-jest"
+      }
```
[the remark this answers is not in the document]
stephtr: Thanks, I added those two snippets. Additionally I separated the code into two different `DebugConfigurationProvider`, such that it shows up as separate environment when creating the launch.json file.

package.json
```diff
@@ -123,6 +123,49 @@
...
+    ],
+    "debuggers": [
+      {
+        "type": "node",
+        "label": "Debug Jest tests using vscode-jest"
+      },
+      {
+        "type": "vscode-jest-tests",
+        "label": "Debug Jest tests using vscode-jest",
+        "languages": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
```
stephtr: I would prefer to target only `*.test.(j|t)sx?` files, but when creating a separate language definition, it seems like I can't base it on the javascript(react)/typescript(react) definitions (including syntax highlighting, IntelliSense, ...).

stephtr:
Should be use the `pathToJest` and `pathToConfig` configuration settings when creating a debug configuration or falling back to the default one?

In the long run my whish would be to completely remove those settings and instead in `function pathToJest(...)` (helpers.ts) parse the launch configuration.

package.json
```diff
@@ -123,6 +123,49 @@
         "command": "io.orta.jest.coverage.toggle",
         "title": "Jest: Toggle Coverage Overlay"
       }
+    ],
+    "debuggers": [
+      {
+        "type": "node",
+        "label": "Debug Jest tests using vscode-jest"
+      },
```
connectdotz: don't think we need this dummy one anymore...

package.json
connectdotz: I think that should be fine for now, if it's a platform limitation...

package.json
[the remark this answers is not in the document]
connectdotz: hmm... I like the snippets, but not sure about splitting DebugConfigurationProvider... I think the original single DebugConfigurationProvider is cleaner. These snippets should just work with your original DebugConfigurationProvider registered under "node" type, right? As we discussed earlier, the `vscode-jest-tests` is really a **fake** type, our DebugConfigurationProvider is really a node type debug provider, IMHO, better to keep that clarity in code.

src/JestExt.ts
```diff
@@ -562,7 +459,19 @@ export class JestExt {
...
-    vscode.debug.startDebugging(vscode.workspace.workspaceFolders[0], configuration)
+    const workspaceFolder = vscode.workspace.workspaceFolders[0]
+    try {
+      // try to run the debug configuration from launch.json
+      await vscode.debug.startDebugging(workspaceFolder, 'vscode-jest-tests')
+    } catch {
+      // if that fails, there (probably) isn't any debug configuration (at least no correctly named one)
+      vscode.window.showWarningMessage(
+        'Debug Jest test: No correctly named debug configuration found in launch.json, starting debugging using the default configuration.'
+      )
```
connectdotz: no need to show warning here as it is the default behavior when the user doesn't have any custom debug config, which is not required.

src/JestExt.ts
```diff
@@ -562,7 +459,19 @@ export class JestExt {
...
+      // therefore run the test using the default configuration
+      const debugConfiguration = this.snippetDebugConfigurationProvider.provideDebugConfigurations(workspaceFolder)[0]
+      await vscode.debug.startDebugging(workspaceFolder, debugConfiguration)
```
connectdotz: this `startDebugging` could fail because the generated config is not correct, for example, the program or runtimeExecutable is wrong... therefore, a good place to catch and instruct users how to examine and customize their debug config.

package.json
```diff
@@ -123,6 +123,49 @@
...
+              "runtimeExecutable": "${workspaceFolder}/node_modules/.bin/react-scripts",
+              "args": ["test", "--runInBand"],
+              "cwd": "${workspaceFolder}",
+              "console": "integratedTerminal",
+              "protocol": "inspector",
+              "internalConsoleOptions": "neverOpen"
+            }
+          }
+        ]
+      }
```
connectdotz: hmm... I like the snippets, but not sure about splitting DebugConfigurationProvider... I think the original single DebugConfigurationProvider is cleaner. These snippets should just work with your original DebugConfigurationProvider registered under "node" type, right? As we discussed earlier, the `vscode-jest-tests` is really a **fake** type, our DebugConfigurationProvider is really a node type debug provider, IMHO, better to keep that clarity in code.

package.json
stephtr: I initially thought that it would be necessary when registering a `CustomDebugProvider`, but it turned out it isn't, I therefore removed it.

src/JestExt.ts
stephtr: I added it because of your "maybe we can capture error here as well and provides some helpful notes to improve the user experience…", but see the next comment.

src/JestExt.ts
stephtr: You mean the "fallback" `startDebugging`? The issue is that even in such cases `startDebugging` resolves the Promise successfully, after automatically displaying an error message (`The attribute "runtimeExecutable" doesn't exist ("/path/to/jest").`). One therefore has to initiate a change of the VS Code API or show the user a warning while falling back to the default configuration or just let VS Code show its error message.

src/extension.ts
```diff
@@ -52,6 +52,11 @@ export function activate(context: vscode.ExtensionContext) {
     ),
     vscode.commands.registerCommand(`${extensionName}.run-test`, extensionInstance.runTest),
     vscode.languages.registerCodeLensProvider(languages, extensionInstance.debugCodeLensProvider),
+    vscode.debug.registerDebugConfigurationProvider('node', extensionInstance.nodeDebugConfigurationProvider),
+    vscode.debug.registerDebugConfigurationProvider(
+      'vscode-jest-tests',
+      extensionInstance.snippetDebugConfigurationProvider
```
connectdotz: you could just register the `extensionInstance.nodeDebugConfigurationProvider` here instead

src/DebugConfigurationProvider.ts
```diff
@@ -0,0 +1,80 @@
...
+        debugConfiguration.args.push(this.testToRun)
+      }
+      this.fileNameToRun = ''
+      this.testToRun = ''
+    }
+    return debugConfiguration
+  }
+}
+
+export class SnippetDebugConfigurationProvider implements vscode.DebugConfigurationProvider {
```
connectdotz: > Having a single DebugConfigurationProvider would be cleaner, for sure. However we need the vscode-jest-tests provider such that the Debugconfigurationprovider appears as a separate entry when initially creating a launch.json config and selecting the environment.

you can accomplish this via registering the same DebugConfigurationProvider under both "node" and "vscode-jest-tests" (see comments in `extension.ts`), without splitting the DebugConfigurationProvider...

src/DebugConfigurationProvider.ts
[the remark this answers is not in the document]
connectdotz: cool

package.json
connectdotz: > Having a single DebugConfigurationProvider would be cleaner, for sure. However we need the vscode-jest-tests provider such that the Debugconfigurationprovider appears as a separate entry when initially creating a launch.json config and selecting the environment.

you can accomplish this via registering the same DebugConfigurationProvider under both "node" and "vscode-jest-tests" (see comments in `extension.ts`), without splitting the DebugConfigurationProvider...

stephtr: But then it doesn't only add the config when creating the launch configuration and selecting `Debug Jest tests using vscode-jest`, but also when selecting `Node.js` instead.

stephtr:
I updated the PR such that it adds `jest.pathToConfig` to the generated debug configuration.
Converting `jest.pathToJest` on the other side isn't that straight forward, since we would have to somehow differentiate between it being a `program` or a `runtimeExecutable` and often running the widespread `npm test --` isn't enough for debugging (best example is create-react-app).

package.json
connectdotz: true, but not sure why is that a problem...

stephtr: I found it somehow unesthetic, but since this is only the case when the extension is activated (and therefore jest is in use), I just merged them.

connectdotz:
I am not quite sure we should add any of these existing settings at the initial implementation... I think people are using them mainly for non-debug mode, which might not be compatible or desirable during debugging, for example, it could be configured with extra params, so 
1. you had to add extra complexity to parse and maybe more 
1. it might not even be desired for debugging purpose, so do you create yet another setting for debug or ...? 
1. I am not sure if these can be added to the snippets, otherwise, they will be out of sync, which is not ideal

Giving we really don't know how users will be using this new debug mechanism, I think it is best to keep our initial implementation as simple and clean as possible. So we could release it sooner and start to get real feedbacks ASAP, before committing more time and effort into a direction that we are not quite sure yet...

I really like your initial implementation (from the first review), I thought that was a balanced approach: that we provide a **simple minimal config**, and users can customize it when the default one doesn't work for them -- simple and intuitive. I thought we were pretty close at review 1, just needed to add the snippets in the package.json, fix the compile error and tests, add a documentation then it should be ready to go...

Other than the doc, I think you already fixed most mentioned above. I suggest we hold off the extra logic outside of that scope, so we can get this out as soon as possible. We can always come back to it later when there are more concrete use cases and actual user feedbacks...

package.json
connectdotz: 👍

stephtr:
> I think people are using them mainly for non-debug mode

Ah ok, I wasn't sure about that and since including `pathToConfig` was the old behavior when debugging a test (and there even was a test about that) I previously added that functionality.

I think except documentation (beyond this PR) everything should be fixed by now.

tests/extension.test.ts
```diff
@@ -92,5 +92,11 @@ describe('Extension', () => {
       expect((vscode.commands.registerCommand as jest.Mock<any>).mock.calls).toContainEqual(expected)
       expect(context.subscriptions.push.mock.calls[0]).toContain(expected)
     })
+
+    it('should register a DebugConfigurationProvider', () => {
+      activate(context)
+
+      expect(vscode.debug.registerDebugConfigurationProvider).toBeCalled()
```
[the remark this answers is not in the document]
stephtr: I don't think that we would gain much. TypeScript already enforces that we actually register a `DebugConfigurationProvider`, so in my opinion further specifying the test would be more a sort of burden when changing the behavior than prevent inadvertent changes.

stephtr:
It should work with the unpublished state of `jest-editor-support`, however it also works fine with `"jest-editor-support": "22.4.0"`. I tried to commit that, but unfortunately prettier fails to run as soon as there is a change in my package.json.

package.json
```diff
@@ -123,6 +123,45 @@
...
+        "languages": ["javascript", "javascriptreact", "typescript", "typescriptreact"],
+        "configurationSnippets": [
+          {
+            "label": "Jest: Default jest configuration",
+            "description": "Running tests by directly running jest",
+            "body": {
+              "type": "node",
+              "name": "vscode-jest-tests",
+              "request": "launch",
+              "program": "${workspaceFolder}/node_modules/jest/bin/jest",
```
connectdotz: I tested adding these snippets in launch.json, all of them failed to start debugger... looks like the `${workspaceFolder}` is not populated. Looking at the example from [here](https://code.visualstudio.com/docs/extensions/example-debuggers), you will most likely need to escape them, for example: `"program": "^\"\\${workspaceFolder}/node_modules/jest/bin/jest\""`

were you able to run with these snippets?

tests/extension.test.ts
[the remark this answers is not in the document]
connectdotz: this is to protect others from making unintended changes, such as removed registration for "vscode-jest-tests". Since we don't have any other tests for launch-snippets function, I think these simple tests are the least we can do to safeguard its intended logic. 

All tests are creating friction for future changes, and that is not necessary a bad thing...

package.json
stephtr: Ah thank you, since they should have been the same, I actually never tested the snippets from `package.json` except with my eyes; is fixed.
I also extended the test and the readme file.

package.json
```diff
@@ -123,6 +123,45 @@
...
+          {
+            "label": "Jest: Default jest configuration",
+            "description": "Running tests by directly running jest",
+            "body": {
+              "type": "node",
+              "name": "vscode-jest-tests",
+              "request": "launch",
+              "program": "^\"\\${workspaceFolder}/node_modules/jest/bin/jest\"",
+              "args": ["--runInBand"],
+              "cwd": "${workspaceFolder}",
```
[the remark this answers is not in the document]
stephtr: Oh, I overlooked that since it also works without a `cwd`. However including the `cwd` has the advantage of better directing the user if he has to change paths.
Debugging of `react-scripts-ts` seems to be not working at the moment with newer versions of jest ((wmonk/#276)[https://github.com/wmonk/create-react-app-typescript/issues/276]).

README.md
```diff
@@ -35,6 +35,14 @@ If have a more complex setup, it can probably be supported, check out the settin
...
 Also, you should use Jest 17+, however 16 works - it will just offer a warning. We're aiming to try and do current Jest version - 1, but won't specifically drop support for older versions unless we're forced into it.
 
+## How do I debug tests?
+
+The simplest use cases should be supported out-of-the-box, but at the latest when VS Code displays errors about the attribute `program` or `runtimeExecutable` not being available, you have to create your own debug configuration within `launch.json`.
+
+This plugin provides blueprints for debugging plain Jest setups or projects bootstrapped by `create-react-app`. If those don't match your setup, you can modify the blueprints or create a completely new debug configuration, but keep in mind, that the `type` has to be `node` and that the configuration has to be named `"vscode-jest-tests"`.
+
+Starting with debugging is possible by clicking on the `debug` CodeLense above appendant `it` tests, but you can also debug all tests at once by starting debugging of `"vscode-jest-tests"` within the VS Code Debug Side Bar.
+
```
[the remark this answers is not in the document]
stephtr: I tried to address that with a separate paragraph. By the way, ejected CRA's were something I didn't thought about, but adding a working snippet was quite simple.

You don't have to feel sorry for that, with each time the quality of this PR got improved.

src/DebugConfigurationProvider.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    const testCommand = folder && getTestCommand(folder.uri.fsPath)
+    if (isCRATestCommand(testCommand)) {
+      const craCommand = testCommand.split(' ')
+      // Settings specific for projects bootstrapped with `create-react-app`
+      debugConfiguration.runtimeExecutable = '${workspaceFolder}/node_modules/.bin/' + craCommand.shift()
+      debugConfiguration.args = [...craCommand, ...debugConfiguration.args]
+      debugConfiguration.protocol = 'inspector'
+    } else if (testCommand === 'node scripts/test.js --env=jsdom') {
```
connectdotz: me: _"uh-oh, this is fragile..."_

me: _"yeah, but that is ok because if it failed developers can just override it in their custom debug config"_

me: _"ok, then why do we even need this code here? when developers eject CRA, I think they would expect some adjustment to their environment, such as their debug config; on the other hand, it is not so intuitive that their trusted vscode-jest suddenly failed to debug after they changed the test script in package.json."_

me: _"maybe, but if vscode-jest can work out-of-box in their CRA environment, wouldn't it be "intuitive" to think it should work in ejected env as well? if the testCommand check is too fragile, there are many ways to improve it: we can parse all package.json scripts,  parse the testCommand further, go down to the file system to find the binary, just to name a few; and if different CRA version has different test scripts, not a problem, if we can find CRA binary, we can find its version... there is no problem we can't resolve!"_

me: _"This is what we call the slippery slope... at the beginning, everything looked equally plausible, but as you travel down the path, the effect starts to compound...before you know it, there will be hundreds of lines of code serving CRA variations and yet still insufficient... now just imagine if we have chosen the other path..."_

----

 "Just because we can doesn't mean we should" 😏

connectdotz: > Should I remove it?

sounds good to me.

stephtr: Is removed.

reaktivo:
Hi! I was looking for this exactly. But I just noticed this hasn't been published to the Marketplace. Is there any ETA? Would love to take advantage of it!

orta:
Yeah, let's do it

PR #293, opened 2018-04-02
Start Jest with the --no-color CLI argument

seanpoulter:
This PR will prevent the ANSI escape codes from appearing in the test messages. This is dependent on  a change to `jest-editor-support` that adds the `--no-color` option (facebook/jest#5909).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Bug-fixes within the same version aren't needed
 
 ## Master
 
+* Prevent ANSI escape codes from appearing in test messages - seanpoulter
 * Your message here - name
 
 -->
--- a/src/JestProcessManagement/JestProcess.ts
+++ b/src/JestProcessManagement/JestProcess.ts
@@ -18,8 +18,11 @@ export class JestProcess {
     this.stopRequested = false
     let exited = false
 
-    const useShell = platform() === 'win32'
-    this.runner = new Runner(this.projectWorkspace, { shell: useShell })
+    const options = {
+      noColor: true,
+      shell: platform() === 'win32',
+    }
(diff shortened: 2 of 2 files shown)


orta:
I’ve merged the other, but this now relies on a new jest release

seanpoulter:
We're still waiting for a Jest deployment. It's making me wish we could go bump `jest-editor-support` on it's own.

orta:
There's nearly always a jest release: https://www.npmjs.com/package/jest

stephtr:
Switching to a beta release of `jest-editor-support` (for example 23.0.0-charlie.2) should work (since it doesn't contain any other beside our own changes).

seanpoulter:
I wonder if we can fix this issue now without the option. 🤔

seanpoulter:
> I wonder if we can fix this issue now without the option. 🤔

On second thought it's not worth any thought.

> Switching to a beta release of `jest-editor-support` ... should work

Do we need to worry about breaking anything else? Things aren't very stable right now ... 🔥

seanpoulter:
Was there something about Jest 23 that solved this problem for us? I don't remember ...

connectdotz:
the new option `noColor` depends on the the changes in `jest-editor-support` [jest/5909](https://github.com/facebook/jest/pull/5909) you submitted a while back.

During #341, I briefly tested the noColor option and it seems to work, however I have only tested with jest 23.x. I think as long as it won't break other jest versions, we should be fine.

seanpoulter:
OK, thanks for refreshing my memory. I'll rebase that when I get a few moments.

seanpoulter:
I've rebased the PR which should be good to go @connectdotz.

connectdotz:
thanks @seanpoulter

PR #296, opened 2018-04-04
Confirm we can deploy in CI

seanpoulter:
Just saw your note to self on #287. Is this all we need to do? 😉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,4 @@ script:
   - yarn ci
   - cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js
   - yarn danger
+  - yarn vscode:prepublish


connectdotz:
you wish 😏 ... we should probably also add `tsc -p xxx` as part of the `test` script in package.json so developers can discover the problem before submitting the PR...

seanpoulter:
Does it make sense if we set up a pre-test script to run `tsc -p ./tsconfig.publish.json`?

I haven't tested it yet, but if compiling fails that would block developer testing and in CI. 🤔

orta:
I'm happy to see all that 👍

seanpoulter:
@connectdotz, you've got the big PR to update **package.json** and **CONTRIBUTING.md**. I think this oneline change is safe to merge.

PR #301, opened 2018-04-13
add option jest.cli.noColor for better error message display

Hental:
jest-editor-support Runner now support new Option `noColor` to add arg `--no-color` for jest process,so i add a new option `jest.cli.noColor (default: true)`, it can display error message better

see also: #279 

thanks :)

PR #303, opened 2018-04-16
add option jest.cli.noColor for better error message display

Hental:
#301

### DIFF
--- a/src/JestProcessManagement/JestProcess.ts
+++ b/src/JestProcessManagement/JestProcess.ts
@@ -1,5 +1,6 @@
 import { platform } from 'os'
 import { Runner, ProjectWorkspace } from 'jest-editor-support'
+import { getExtensionCliSettings } from '../extension'
 
 export class JestProcess {
   static readonly keepAliveLimit = 5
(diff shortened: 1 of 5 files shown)


seanpoulter:
Sorry for the delay replying to your question about the mocked modules. Any reason for closing the PR?

Hental:
because i find the test error caused by i call method getExtensionCliSettings to get cli settings in class JestProcess, and i realize maybe i can  expand pluginSettings and get cli setting from pluginSettings, but that means i will pass the argument by many methods, and that maybe isn't a good idea, maybe #293 is simple and better.
thank you for read it, my english is poor.

src/JestProcessManagement/JestProcess.ts
```diff
@@ -19,7 +20,8 @@ export class JestProcess {
     // Use a shell to run Jest command on Windows in order to correctly spawn `.cmd` files
     // For details see https://github.com/jest-community/vscode-jest/issues/98
     const useShell = platform() === 'win32'
-    this.runner = new Runner(this.projectWorkspace, { shell: useShell })
+    const cliSettings = getExtensionCliSettings()
```
Hental: this lin cause test fail

Issue #306, opened 2018-04-21
Not working with typescript files


yarkovaleksei:
### Environment

1. `node -v`: v9.8.0
2. `npm -v`: 5.8.0
3. Operating system: Ubuntu 14.04

### Steps to Reproduce

1. Install extension to VSCode and restart
2. Open `module.spec.ts`
3. Do not work Jest autostart

yarkovaleksei:
@seanpoulter Oh sure. Jest is used to perform tests from the command line.

seanpoulter:
If Jest is installed globally it will not trigger the extension to start. Any chance that's the problem?

As you'll see in the list below that is taken from the [`activationEvents`](https://github.com/jest-community/vscode-jest/blob/b4fb57e9a50217584426244702ea732d29c539ae/package.json#L23) in package.json, Jest will start when:
* the workspace contains any of:
  - `jest.config.js`
  - `jest.json`
  - `node_modules/.bin/jest`
  - `node_modules/react-scripts/node_modules/.bin/jest`
  - `node_modules/react-native-scripts`
* the user runs the **Jest: Start Runner** command

Does that help?

yarkovaleksei:
My file is here: `test/unit/jest.config.js`.
Could this be a problem?

seanpoulter:
That's the problem. We're only checking at the root level because Jest only checks at the root level unless you provide it a CLI option.

PR #307, opened 2018-04-21
Add documentation for the extension activation events

seanpoulter:
Here's some proposed documentation based on a reply to Issue #306.


orta:
Yep 😀

Issue #310, opened 2018-04-24
Don't see coverage info in files being tested


garyking:
I've already got coverage info being created by Jest, yet I don't see coverage info for files that are being tested. 

Where am I supposed to see this info? In the gutters?

seanpoulter:
Does this help? I've just proposed a change to the docs.

<img width="918" alt="screen shot 2018-04-24 at 08 45 18" src="https://user-images.githubusercontent.com/2585460/39187666-e84f73be-479b-11e8-9326-8f8d250b2e71.png">

seanpoulter:
Here's an example of what the coverage overlay looks like taken from another issue:

<img alt="Coverage overlay example" src="https://user-images.githubusercontent.com/2585460/35473903-9bb7aad4-0354-11e8-8fda-e3160486c8b7.gif">

garyking:
Thanks, but when I run `Jest: Toggle Coverage Overlay`, nothing happens. 

I create Jest and open coverage reports with the following in `package.json`:

```json
"scripts": {
  "test": "jest __tests__ --coverage",
  "coverage": "open coverage/lcov-report/index.html"
}
```

garyking:
For some reason, it works now.

chenxiaochun:
@garyking , i am get same issue and i have been tried a variety of way and  still don't know the reason too.

PR #302, opened 2018-04-14
Improving coverage display? #285 => Add GutterFormatter

Guymestef:
Here a first implementation using gutters

PR #312, opened 2018-04-24
Add code coverage to documenation

seanpoulter:
Adding more to the docs to help with issues like #310.

I've blocked out the docs that we should add as #302 gets closer to shipping.


seanpoulter:
Thanks @orta! 👏

PR #323, opened 2018-05-11
Highlight the full line of the jest error, not just the first 6 characters.

ThomasRooney:
It bothered me that just the first 6 characters of each line were highlighted for the erronous jest diagnostic statement.

This makes it highlight the full line instead.

If the editor is open, it highlights from the trimmed firstNonWhitespaceCharacter, otherwise it's highlighting the full line. This means the columns highlighted swap slightly when the file is opened. I couldn't see a way around this but for either parsing the error message structure from jest, or loading the file into memory; neither nice options. 

Old:

old-vscode-error

New:

new-vscode-error

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ Bug-fixes within the same version aren't needed
 
 ## Master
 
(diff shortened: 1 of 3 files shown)


orta:
Nice one, yeah, this looks 👍  to me

PR #326, opened 2018-05-13
Fix tests with same name corrupting the decorator dots.

ThomasRooney:
Two tests with the same name (e.g. in different describe blocks) currently get the decorator of the last test result.

This is bad behaviour, as its common practice when testing an API under different conditions, to have the scenario be in the `describe` block and the API method under test in the `test` block. I.e. its a valid scenario to have multiple `test` blocks with the same key.

This adjusts that behaviour, by using the line numbers of the test under KnownFail scenarios to assign the decorators correctly, or otherwise just using the first Success or Unknown case (as they have the same parsed test result anyway -- line numbers being only available in KnownFail scenario).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Bug-fixes within the same version aren't needed
 
 ## Master
 
+* Fix decorators showing corrupt values when two tests of the same name in one file (e.g. in different describe blocks) exist - ThomasRooney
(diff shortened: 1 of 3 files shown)


orta:
Another fabulous PR, great job - I've sent you an invite to admin this repo. 

I'll get these shipped.

orta:
Shipped as 2.7.2

PR #328, opened 2018-05-20
Support commands without active text editor

### LINKED ISSUE #318: "Jest: Start Runner" only works when a file is open
I have `jest.autoEnable` set to `false`. When I open a workspace with Jest installed, and I run `Jest: Start Runner`, nothing happens. If I open a file then run it, then it works fine.

So, it doesn't seem to run if there is no file open. It doesn't matter if the file that's open is a test file or not, either.

seanpoulter:
This issue will close #318. 

Changes proposed in this PR:
* It was reported that the **Jest: Start runner** command would not work when there wasn't an active text editor open. Changing the call from `registerTextEditorCommand` to `registerCommand` fixes this.
  
  The change has been tested locally and works OK.
  There are no changes to the tests. 😓

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ Bug-fixes within the same version aren't needed
 ## Master
 
 * Adds a setting to control when the debug CodeLens appears - seanpoulter
+* Support the "Jest: Start/Stop" and "Show output" commands without an active
+  text editor - seanpoulter
 
 -->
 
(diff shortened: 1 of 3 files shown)


seanpoulter:
For what it's worth, sorry for the new development. I've been trying to get a sense of what regressions we've introduced and noticed this is really really low hanging fruit. 🍒

orta:
Seems 👍  to me, yep

orta:
No apologies, do whatever you want to work on, anyone who is using this tool has the abilities to contribute to it. Oeople making issues or requesting features are just choosing to prioritize other things, you can too  👍

PR #329, opened 2018-05-20
When --watch is not supported, restart with --watchAll

### LINKED ISSUE #327: All tests re-run on every change
### Environment

1. `node -v`: 8.11.1
2. `npm -v`: 5.6.0

3. Operating system: macOS High Sierra

### Steps to Reproduce

Save a change to any specific test.

### Expected Behavior

Only the changed file's tests are re-run, leading to quick feedback on test success/failure as a result of the change.

### Actual Behavior

All tests in the entire project are re-run, potentially taking a very long time before finding out if the one modified test passes/fails.

This appears to be a regression due to #325 . As one who didn't have problems with the previous behavior, it would be very nice if this could at least be made an optional setting for those who need it.

seanpoulter:
The change introduced in #325 incorrectly configured the Runner to start with `--watchAll` enabled. I'll take some of that blame for the unclear API. To use `--watchAll` you need to set `true, true` which isn't clear. 😞 

Changes proposed in this PR:
* Change the watch mode flag from a Boolean to an enumeration: none, watch, watchAll
* Properly start the Runner with `--watch` and `--watchAll`
* Detect the error message when `--watch` is not supported and restart with `--watchAll`

Changes not proposed in this PR:
* No setting to "always use --watchAll" since we can detect when `--watch` fails
* Moving the error message detection into `jest-editor-support`. I know **cpojer** wants to add integration tests to make sure we catch if the text changes.

Testing:
* Manually tested using a workspace without a git or hg repo:
  
  **index.test.js**:
  ```js
  describe('Test suite', () => {
    it('Test', () => {
      expect(1).toBe(2);
    })
  })
  ```
* Tests have been updated with a few redundant test cases removed

This will resolve #327.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Bug-fixes within the same version aren't needed
 * Adds a setting to control when the debug CodeLens appears - seanpoulter
 * Support the "Jest: Start/Stop" and "Show output" commands without an active
   text editor - seanpoulter
+* Restart Jest with --watchAll when --watch is not supported without git/hg
+  - seanpoulter
 
 -->
 
--- a/src/Jest/index.ts
+++ b/src/Jest/index.ts
@@ -1,0 +1,8 @@
+export enum WatchMode {
+  None = 'none',
+  Watch = 'watch',
+  WatchAll = 'watchAll',
+}
+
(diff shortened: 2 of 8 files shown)


orta:
👍 cool, yeah, makes sense

I'll ship a release

Issue #226, opened 2018-01-16
Coverage toggling


smith:
It would be nice to be able to toggle coverage checking (with `--coverage`/`--no-coverage`.) 

Toggling the overlay is useful, but the slowest part of my test run is generating the coverage report. I'd like to be able to toggle this quickly to where I could normally leave it off but turn it back on when wanting to do coverage checking.

seanpoulter:
Brilliant. We'd love your help getting that implemented @smith. I'm happy to help point out the code that we'll want to change. It'd be great to have a second set of eyes on the coverage and process restarting code as well -- both have some changes on the go.

stephtr:
In my opinion adding a new vscode setting (like `jest.collectCoverage`) which can be flipped using a command would be the best way to achieve that. That way one also doesn't have to manually add `jest.pathToJest` if its default value works.

smith:
facebook/jest#5836 adds what I think is needed it implement this.

connectdotz:
I can see 2 coverage usage patterns:
- **on-demand**: like @smith mentioned, an occasional jest coverage run is sufficient, a command to toggle on/off both display and an 'one-off' jest run with coverage flag will be good enough.
- **continuous**: as users add more tests, they might want to see the coverage updated as each jest run. In this case, users probably should update their regular jest command line to append the `--coverage` flag, but we should still have a command to toggle the display in case the coverage is somewhat impacting development visual (my personal experience with code coverage display)...

In any case, I think the following logic should be able to cover both:
- have a command to toggle coverage: mainly to turn on/off coverage UI display. Optionally: it could be enhanced with keyboard shortcut or on-screen control for ease of use.
- if the display coverage command is "on" but we don't have the coverage report to show for, i.e. coverage report is not in the jest output: start a jest run with coverage flag with [facebook/jest#5836](https://github.com/facebook/jest/pull/5836) `coverage` option. If users already have the `--coverage` in their rgular jest command line as for contineous coverage reporting, i.e. we can find coverage report: we can omit this step all together.

@stephtr do you see any problem with this logic? If users can run regular jest without modifying `jest.pathToJest`, the runner ([facebook/jest#5836](https://github.com/facebook/jest/pull/5836)) should be able to just append the `--coverage` flag without any new setting, wouldn't it?

stephtr:
[...]

PR #330, opened 2018-05-21
Add a setting for collecting coverage

stephtr:
[Jest/#5836](https://github.com/facebook/jest/pull/5836) enabled switching whether or not Jest should collect coverage information by a workspace attribute. This PR adds a setting to control that attribute.

Points to talk about:

- I defaulted the (boolean) setting to `null`, since supplying `false` would add the `--no-coverage` flag.
- Toggling coverage overlay may not display the overlay if `collectCoverage` isn't `true`, it therefore shows a warning message in that case. That behavior isn't perfect if `collectCoverage` has been kept at `null` and the jest command/config has been modified to nevertheless collect the information.
- The same is true for `showCoverageOnLoad`, but without a warning message.

An alternative would be that, additional to the `coverageCollect` switch, the setting could be temporarily overwritten when toggling the coverage overlay.

By the way: Sorry that I accidentally pushed these commits to the master branch, that shouldn't have happened, but I had my local branch tracking upstream instead of origin.


stephtr:
One concrete use case would be if the default value of `pathToJest` would be fine for a project, one doesn't have to manually add the path to jest.
We furthermore don't force the user to use the new setting (maybe I should remove the warning message), he could still specify the `--coverage` flag within `pathToJest`.
I think it's generally cleaner to separate the path setting from the executable's extra arguments (like `pathToConfig`, which could be also supplied via).

connectdotz:
> I think it's generally cleaner to separate the path setting from the executable's extra arguments (like pathToConfig, which can be also supplied via a setting or as a flag).

I like this idea a lot better, and I think it will probably give us more bang for the buck in the long run...  But can you give us a little bit more background on what exactly the problem we are trying to solve with this PR?

stephtr:
There's no problem with the current status, it would be just a simplification/enhancement. But maybe it would be better to continue the discussing in #226.

connectdotz:
oh i see, will do

seanpoulter:
I prefer having folks configure Jest separate from configuring the UI. Jest has great documentation. If anything, I’d want to help change the config for users.

stephtr:
In what direction would you then like #226 to go?

seanpoulter:
I'm not sure @stephtr and while it's a good conversation to have I'd urge us to focus on our core functionality instead of the nice to have issues. While trying to fix a few regressions we've introduced in the last few weeks I seen all of these:
* test decorators not updating properly
* inline error messages that aren't being cleared
* colour escape codes appearing in messages
* the status bar message that hangs
* inline error messages in the wrong spot

I'll skim #226 and give my two cents.

Issue #313, opened 2018-04-24
Test suite failed to run in VSCode 1.22.2


karb0f0s:
### Environment

1. `node -v`: 9.11.1
2. `npm -v`: 5.6.0
3. `npm ls react-scripts` (if you haven’t ejected): 
no
4. Operating system: 
Microsoft Windows [Version 10.0.14393]
5. `VSCode`: version 1.22.2

### Steps to Reproduce

I was able to reproduce error with this repo https://github.com/karb0f0s/vscode-jest-test/ on Windows. Although this error doesn't appear on MacOS. I guess it is somehow related to the way VSCode spawns `jest` process.

package.json
```json
{
  "name": "app",
  "private": true,
  "scripts": {
    "test": "jest",
    "test:watch": "jest --watch",
    "build": "webpack",
    "watch": "webpack --watch"
  },
  "devDependencies": {
    "@babel/cli": "^7.0.0-beta.46",
    "@babel/core": "^7.0.0-beta.46",
    "@babel/preset-env": "^7.0.0-beta.46",
    "babel-core": "^7.0.0-bridge.0",
    "babel-jest": "^22.4.3",
    "babel-loader": "^8.0.0-beta.2",
    "jest": "^22.4.3",
    "webpack": "^4.6.0",
    "webpack-cli": "^2.0.15"
  },
  "jest": {
    "verbose": true,
    "bail": true,
    "testRegex": "(/__tests__/.*|(\\.|/)(test|spec))\\.(js|ts)$",
    "transform": {
      "^.+\\.js?$": "babel-jest"
    }
  }
}
```

.babelrc
```json
{
  "presets": [
    ["@babel/preset-env", {
      "targets": {
        "browsers": ["last 2 versions", "ie >= 10"]
      },
      "modules": false,
      "useBuiltIns": "usage",
      "debug": true
    }]
  ],
  "env": {
    "test": {
      "presets": ["@babel/preset-env"]
    }
  }
}
```

### Expected Behavior

`npm test` works as expected, `Debug` under the `test` in source file works as expected, `Debug configuration` `vscode-jest-tests` works as expected, with transpilation and all

### Actual Behavior

VSCode Jest plugin fails to run. Output contains:
```
FAIL test\countPrice.test.js

  ● Test suite failed to run

    .\app\test\countPrice.test.js:1
[...]

PR #297, opened 2018-04-05
Add usage of shell when creating Settings

stephtr:
This change will finally close #274. We still have to wait until [jest/#5658](https://github.com/facebook/jest/commit/96d5453f62bd317353d1d19364d9e659ccc5aa40) and [jest/#5872](https://github.com/facebook/jest/commit/358764dcc784028e5e2d231a0e40e02688fc532b) get released to npm, but in the meantime I created this PR so that we don't forget about it.

PR #332, opened 2018-05-22
Fix regression detecting create-react-app

seanpoulter:
As discussed in Issue #313 and the comments on PR #324, we've introduced a regression that broke support for create-react-app out of the box.

Changes proposed in this PR:
* This changes when we check if the workspace was bootstrapped with create-react-app. We've recently changed the default value for `pathToJest` to `null` so the following check would only be performed if the user has changed their settings. This caused new bootstrapped projects to run with `jest` instead of indirectly calling `react-scripts test` via `npm test --`.

    The following caused the regression:
    ```js
       if (pluginSettings.pathToJest) {	
        if (isBootstrappedWithCRA(pluginSettings.rootPath)) {	
    ```

Testing plan:
* Tested on OSX without issue
* Added tests for `pathToJest` to capture recent changes including defaulting to global Jest


tests/helpers.test.ts
```diff
@@ -118,10 +129,87 @@ describe('ModuleHelpers', () => {
...
+
+      expect(pathToJest(defaultSettings)).toBe('npm test --')
+    })
+
+    it('returns the normalized "pathToJest" setting when set by the user', () => {
+      const expected = {}
+      mockNormalize.mockReturnValueOnce(expected)
+
+      const settings: any = {
+        pathToJest: 'jest',
```
stephtr: It's just a small thing, but maybe we should test for a different value since `jest` is the default outcome of the `pathToJest` function.

src/helpers.ts
```diff
@@ -61,24 +61,31 @@ function hasNodeExecutable(rootPath: string, executable: string): boolean {
...
-    return normalize(join(pluginSettings.rootPath, 'node_modules', '.bin', 'jest'))
+  const localJestExecutable = pathToLocalJestExecutable(pluginSettings.rootPath)
+  if (existsSync(localJestExecutable)) {
+    return localJestExecutable
   }
+  return `jest${isWindows() ? '.cmd' : ''}`
+}
+
+function isDefaultSettingForPathToJest(str) {
+  return str === null
```
stephtr: It practically doesn't make any large difference, but wouldn't it be nicer if we change the default value of `jest.pathToJest` within package.json to `""` instead of `null`, like `jest.pathToConfig` and `jest.rootPath`?

seanpoulter: I'm going to call you out on a pattern in your comments @stephtr. I find it frustrating when you structure your feedback as "wouldn't it be better if ..." or "wouldn't it be nicer if ..." and don't explain why you favour your suggestion. If you provided an explanation I'd be willing to discuss the options. Without one it comes across to me like you know better, and I should too.

For these default values feel free to change them to whatever you want in another PR. It doesn't make any difference to me. I'd lean towards not changing the default JavaScript behaviour and have values default to `undefined` when they're not provided, or maybe have a look if VS Code suggests any other conventions with their configuration API.

src/helpers.ts
```diff
@@ -61,24 +61,31 @@ function hasNodeExecutable(rootPath: string, executable: string): boolean {
  * @returns {string}
  */
 export function pathToJest(pluginSettings: IPluginSettings) {
+  if (isDefaultSettingForPathToJest(pluginSettings.pathToJest) && isBootstrappedWithCRA(pluginSettings.rootPath)) {
```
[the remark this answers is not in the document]
seanpoulter: > If you really would like to introduce a function for essentially checking if the string is empty

It might seem like a trivial thing to make a function for but the function adds value by:
1. clearly communicating what we're doing
2. isolating the default value so it's easier to change
3. providing a seam to test if we're checking the default value properly

A null check often doesn't provide the context that someone needs to understand where the value came from or what the value means. This can be fixed using (something like a Null Object or [Special Case](https://martinfowler.com/eaaCatalog/specialCase.html)), a comment, or a well named variable or function.

I'd like to help build a community that's engaged and empowered to contribute to this extension. It starts with little things like the trivial function or advocating for code that's easier to maintain in the long run. It may be more verbose but it's a small overhead now that should help us all maintain it in the long run.
<br/>

> four lines later you also should use that function

Do you mean to check if the user has changed the value in `if (pluginSettings.pathToJest)`? That's reordered and changed it to `if (isNotDefaultPathToJest(pathToJest))`.
<br/>

> move the if(pluginSettings.pathToJest) up to the top

Great idea 👍.

src/helpers.ts
```diff
@@ -104,9 +111,9 @@ export function pathToJestPackageJSON(pluginSettings: IPluginSettings): string |
     pathToNodeModules = join(pluginSettings.rootPath, relativePathToNodeModules)
   }
 
-  const defaultPath = normalize(join(pathToNodeModules, 'jest/package.json'))
-  const cliPath = normalize(join(pathToNodeModules, 'jest-cli/package.json'))
-  const craPath = normalize(join(pathToNodeModules, 'react-scripts/node_modules/jest/package.json'))
+  const defaultPath = normalize(join(pathToNodeModules, 'jest', 'package.json'))
```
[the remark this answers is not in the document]
seanpoulter: Thanks. I didn't know that. I've reverted those changes.

tests/helpers.test.ts
seanpoulter: Good call. I've switched it to a `{}`.

src/DebugConfigurationProvider.ts
```diff
@@ -63,7 +63,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
     }
 
     const testCommand = folder && getTestCommand(folder.uri.fsPath)
-    if (isCRATestCommand(testCommand)) {
+    if (isCreateReactAppTestCommand(testCommand)) {
```
seanpoulter: All this talk about improving maintainability made me remember just how long it took to figure out what "CRA" was from reading the issues. Let's spell it out for folks.

src/Settings/index.ts
```diff
@@ -16,3 +16,11 @@ export interface IPluginSettings {
...
   showCoverageOnLoad: boolean
 }
+
+export function isDefaultPathToJest(str) {
+  return str === null
+}
+
+export function isNotDefaultPathToJest(str) {
+  return !isDefaultPathToJest(str)
+}
```
seanpoulter: I've extracted the trivial function and it's complement to join the settings interface.

src/helpers.ts
```diff
@@ -56,29 +56,32 @@ function hasNodeExecutable(rootPath: string, executable: string): boolean {
...
-export function pathToJest(pluginSettings: IPluginSettings) {
-  if (pluginSettings.pathToJest) {
-    if (isBootstrappedWithCRA(pluginSettings.rootPath)) {
-      return 'npm test --'
-    }
-    return normalize(pluginSettings.pathToJest)
+export function pathToJest({ pathToJest, rootPath }: IPluginSettings) {
+  if (isNotDefaultPathToJest(pathToJest)) {
+    return normalize(pathToJest)
   }
```
seanpoulter: Moved as suggested @stephtr. ☝️ Thanks.

seanpoulter:
Does that address your concerns @stephtr?

src/helpers.ts
```diff
@@ -61,24 +61,31 @@ function hasNodeExecutable(rootPath: string, executable: string): boolean {
...
+  return `jest${isWindows() ? '.cmd' : ''}`
+}
+
+function isDefaultSettingForPathToJest(str) {
+  return str === null
+}
+
+function pathToLocalJestExecutable(rootDir) {
+  return normalize(join(rootDir, 'node_modules', '.bin', `jest${isWindows() ? '.cmd' : ''}`))
```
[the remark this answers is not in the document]
stephtr: My approach was to update the `pathToJest` function and add the missing `useShell` flag all in one go (which would eliminate the need for the `.cmd` extension), but that's not necessary, separating it into two PRs also has its advantages.

src/helpers.ts
stephtr: [Your argument](https://github.com/jest-community/vscode-jest/pull/332#discussion_r190103335) is good, I just believe that `if (pathToJest)` ("if pathToJest is set") would be (even for outsiders) more easily readable and understandable than `if (isNotDefaultPathToJest(pathToJest))`.

src/helpers.ts
```diff
@@ -56,29 +56,32 @@ function hasNodeExecutable(rootPath: string, executable: string): boolean {
...
-  const platform = process.platform
-  if (platform === 'win32' && existsSync(join(pluginSettings.rootPath, 'node_modules', '.bin', 'jest.cmd'))) {
-    return normalize(join(pluginSettings.rootPath, 'node_modules', '.bin', 'jest.cmd'))
-  } else if (
-    (platform === 'linux' || platform === 'darwin') &&
-    existsSync(join(pluginSettings.rootPath, 'node_modules', '.bin', 'jest'))
-  ) {
-    return normalize(join(pluginSettings.rootPath, 'node_modules', '.bin', 'jest'))
+  if (isBootstrappedWithCreateReactApp(rootPath)) {
+    return 'npm test --'
```
stephtr: Shouldn't we here also add the `.cmd` file extension on Windows? (at least until #297 hopefully solves all extension related issues)

tests/helpers.test.ts
```diff
@@ -116,12 +128,89 @@ describe('ModuleHelpers', () => {
...
+            test: 'react-scripts test',
+          },
+        })
+      )
+
+      expect(pathToJest(defaultSettings)).toBe('npm test --')
+    })
+
+    it('returns the normalized "pathToJest" setting when set by the user', () => {
+      const expected = {}
```
stephtr: I find it more complicate to use an object and mock `normalize` than a custom string like `'customjest'`, but that's just my personal opinion and nothing which has to be changed.

stephtr:
Sorry for the delay, unfortunately I had an even busier week than usually.

I have only one concern, the [explicit `null` check](https://github.com/jest-community/vscode-jest/pull/332/files#diff-132c564a5d957ba0bb5c3e6eeec3f213R22) in `isDefaultPathToJest`.
Not that long ago the default value for `jest.pathToJest` was `""`, it could therefore be possible that some users have `""` in their local VS Code configurations. The explicit null check would break the plugin for them, since it would then try to execute an empty string.
Apart from that it looks good to me.

stephtr:
@seanpoulter Do you have an ETA or can I somehow assist you?

tests/helpers.test.ts
seanpoulter: It's not great using the wrong type, but since it's an Object that's asserted with `expect().toBe()` you can be confident the value hasn't changed.

src/helpers.ts
seanpoulter: Yes, this is a known problem but that's not what the PR was for. It was intended to be a fast fix. It's the same behaviour as Line 65 on the left, and the same before we've introduced the regression ([here](https://github.com/jest-community/vscode-jest/blob/7964a9d6c675502f8e3f080a3335c975a0908e63/src/helpers.ts#L69)).

src/helpers.ts
seanpoulter: I'll change it to `if (hasUserSetPathToJest(pathToJest))` and call it a day.

seanpoulter:
Good call on supporting the old default `""` value. I don't have any other changes.

The rest is up to you and @orta. ✌️

src/helpers.ts
connectdotz: I bumped into this problem today while testing another PR. I agree #324 was flawed, but to override `jest.pathToJest` without checking if it has already been customized by users will be problematic too, for example, consider for yarn users whose jest.pathToJest was set to `yarn test`, now it will be **silently** override...

If our goal is to provide a "default" setting when `jest.pathToJest` is null, then let's do just that, no more and no less. In general, we should never override users' custom config programmatically (we can suggest, but they should be the one to carry out the change, simply because we could be wrong) and definitely not silently. Actually, giving `jest.pathToJest` is so critical, if we did end up using the default value, I think we should make this information available some where (maybe a new channel: vscode-jest?), we could also use the new channel to communicate the plugin specific info such as workspace info, settings, recommedations etc. 

BTW, this PR has been sitting for a while, what is the plan? People are adopting custom `jest.pathToConfig` that made the regression issue less urgent, but nevertheless we could take this opportunity to improve quality and reduce confusion...

stephtr: > I could pick up your change and test them together (except windows) before we cut the next release…

I would like to also include a PR like #297, such that all pathToJest and Windows problems should be fixed. But that should be a matter of minutes as soon as we can merge this PR.

orta: Let's get them all in, I'll handle the jest ones and get that released in a few days - can we try get those two merged in that timeframe and we'll ship a new release?

stephtr: Sure. Can we merge this one or is there something speaking against?

connectdotz: @stephtr this PR has no `jest-editor-supports` dependency, if you and @seanpoulter all have tested it and are satisfied with it, then it's probably safe to merge, IMHO.

@orta If you can merge [facebook/jest/6586](https://github.com/facebook/jest/pull/6586) and cut a beta jest release, then we can proceed on this end for the PRs depending on the new `jest-editor-supports`.

orta: OK, I've requested a release for editor-support - so let's get this PR in then, and https://github.com/jest-community/vscode-jest/pull/297 can happen whenever

src/Settings/index.test.ts
```diff
@@ -0,0 +1,16 @@
...
+  })
+
+  it('returns true for the legacy default ""', () => {
+    expect(isDefaultPathToJest('')).toBe(true)
+  })
+
+  it('returns false otherwise', () => {
+    expect(isDefaultPathToJest('')).toBe(false)
+  })
+})
```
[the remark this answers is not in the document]
connectdotz: ok, I will move this test into the normal `tests` directory in #341

Issue #337, opened 2018-06-12
create-react-app-typescript displays:'extension relies on Jest 20+ features'


cmenekse:
### Environment

1. `node -v`: **8.11.1**
2. `npm -v`: **5.6.0**
3. `npm ls react-scripts` (if you haven’t ejected): 
it is from create-react-app-typescript so 
**react-scripts-ts@2.16.0**

4. Operating system:  **Windows 10**

### Steps to Reproduce
run 
`create-react-app my-app --scripts-version=react-scripts-ts`
add in vscode settings: 
```
  "jest.pathToJest": "npm run test --",
  "jest.debugCodeLens.showWhenTestStateIn": [
    "fail",
    "pass",
    "skip",
    "unknown"
  ],
  "jest.enableCodeLens": true,
```
then CTRL+SHIFT+P, Jest:Start Runner.

I do not see `jest ` installed globally when doing `npm ls -g --depth 0 `

### Expected Behavior
No Jest 20+ error and dots to appear near tests

### Actual Behavior
Displays : `This extension relies on Jest 20+ features, it will continue to work, but some features may not work correctly`
and no dots are appearing, everything else seems to be fine.

connectdotz:
this indeed is a bug, #341 should fix it once released.

connectdotz:
vscode-jest [2.9.0](https://github.com/jest-community/vscode-jest/blob/master/CHANGELOG.md#290) has shipped with [jest 23.x support and more](#341), which should address this issue. Feel free to reopen if otherwise.

Issue #316, opened 2018-04-28
"Starting watch mode" never finishes & test indicators' state never changes.


zeroEvidence:
Hello,

I'm not sure if I'm the only one experiencing this, but this extension never finishes the "Starting watch mode" phase. Although, all of tests run, but the pass/fail indicator next to the test in the editor never changes from an unfilled green circle to the failed or pass state.

Other than that it's been working very well.

Thanks for a great extension, it's been very helpful, I appreciate the time and work that you've put into it.

Kindest regards.

### Environment

1. `node -v`: v8.11.1
2. `npm -v`: 6.0.0
3. `npm ls react-scripts` (if you haven’t ejected): 
```
@goldberg/nodejs.ts.lib.bare-base@1.0.0 /home/moored/workspaces/au.net.dmdigital.gitlab/goldberg/libraries/bare-base
└── (empty)
```

4. Operating system: 
```
$ cat /proc/version
Linux version 4.10.0-38-generic (buildd@lgw01-amd64-059) (gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) ) #42~16.04.1-Ubuntu SMP Tue Oct 10 16:32:20 UTC 2017

$ lsb_release -a
Distributor ID:	LinuxMint
Description:	Linux Mint 18.3 Sylvia
Release:	18.3
Codename:	sylvia

$ cat /etc/debian_version
stretch/sid
```

### Steps to Reproduce

1) Clone repo: https://github.com/zeroEvidence/bare-base
2) 
```
$ npm install
```
3) Open workspace with vscode v1.22.2, or 1.22.1, or 1.21.1
4) Disable all extensions except for vscode-jest

### Expected Behavior

vscode-jest extension should finish initising watch mode and the test indicators should change state

### Actual Behavior

vscode-jest extension "starting watch mode" runs forever.

### VSCode developer console log
[-1524874876899.log](https://github.com/jest-community/vscode-jest/files/1957064/-1524874876899.log)

zeroEvidence:
Update:
The above behaviour ceased in one of my projects. I am now testing the same project as the repo provided. So far, no luck. But I'll keep it running in the background to see if it changes.

zeroEvidence:
I've figured out that this issue happens if jest is upgraded to one of their early v23 releases. 
Downgrading to the latest v22 resolves this issue.
[...]

PR #341, opened 2018-07-01
jest 23 compatibility and others

connectdotz:
This PR mainly addresses jest 23.x compatibility and missing dots issues:

most of the changes are in [facebook/jest/6586](https://github.com/facebook/jest/pull/6586):
- plugin hung with jest 23.x (#316): 
  - cause: test results messaging has been moved from stdout to stderr in [facebook/jest/4858](https://github.com/facebook/jest/pull/4858)
- CRA ts repo missing dots (#337)
  - cause: `jest-editor-support` Settings failed to extract jest config via `--showConfig` when using CRA test scripts as command: extra output confused JSON parser. In such case, the default jest config will be used, which doesn't recognize typescript test files, therefore no dots displayed.

a few other technical debt, fixes and cleanup within vscode-jest:
- eliminate misleading error message 'extension relies on Jest 20+ features...' when it can't find the jest package.json file, such as in CRA case. 
  - We should only report version issue when we do know that our version is indeed < 20. today since we default to 18 if we failed to retrieve config/version info, it will report as the version issue instead of the actual issue (getSettings() failed).
  - better yet, turns out we don't even need the separate version lookup logic outside of `Setting.getConfig()`. Therefore, removed `getJestVersion` in `JestExt.ts`, but added version/config validation and reporting. 
- remove the duplicate local jest-editor-support typings and defer to the consolidated, official typing from `jest-editor-support` going forward. (looks like #340 has done the same)
- upgrade to the latest jest packages, this means we will pick up all the pending jest-editor-support changes people might be waiting for, such as Snapshot prefix, shell options etc.
- added a new `jest.debugMode` setting to allow more verbose messaging from `jest-editor-support` so developers can self-diagnose problem easier going forward. 
- found a bug in `triggerUpdateSettings()` that not only it didn't refresh jestSettings by calling `getConfig()`, it actually overrides it by  `this.jestSettings = new Settings(this.workspace)`, which will end up using the default jest config instead of the actual one. 

## Testing
- [x] unit tests
- [x] CRA both ts and js flavor
- [x] jest 23.2 and 22, 20 (backward compatibility)
- [x] on mac
- [ ] on windows - **need help here**

## Note

- this PR has dependency on [facebook/jest/6586](https://github.com/facebook/jest/pull/6586).
- (7/1) this PR is not ready for merge - it will not pass the intgration test and will have merge conflict, because we are still waiting for jest release with the PR above and I haven't merged PR #340 and #334 giving there are still active discussions.
- decided to submit the PR prematuretly so we can start the discussion sooner than later, once we settle the issues above, I will update the package.json and merge accordingly.


connectdotz:
@orta how do we find out if jest has cut a release for jest-editor-supports? do we need to file an issue?

connectdotz:
i see, sometimes they cut alpha releases to unblock dependent tools... I wonder if we can raise the urgency to let them know that none of our users can use jest 23.x with this plugin due to the [facebook/jest/4858](https://github.com/facebook/jest/pull/4858), which we fixed in [facebook/jest/6586](https://github.com/facebook/jest/pull/6586) but will need some kind of jest release, even alpha, to unblock.

connectdotz:
ok, all merges and dependencies are up to date now. ready for review...

@stephtr @seanpoulter I don't have windows and I think we should definitely do the eat-out-ow-dog-food testing on windows, can you guys help?

src/JestExt.ts
```diff
@@ -198,22 +202,65 @@ export class JestExt {
...
+    }
 
-    // Do nothing for the minute, the above ^ can come back once
-    // https://github.com/facebook/jest/pull/3592 is deployed
+    // display settings in debug mode
+    if (this.workspace.debug && showSettingPrefix) {
+      console.log(`${showSettingPrefix}:
+                jestVersionMajor=${this.jestSettings.jestVersionMajor}
+                testMatch=${this.jestSettings.settings.testMatch} 
+                testRegex=${this.jestSettings.settings.testRegex}
```
stephtr: The spaces in L244 and L245 before the line endings are probably not intended?

src/extension.ts
```diff
@@ -19,7 +19,15 @@ export function activate(context: vscode.ExtensionContext) {
   const jestPath = pathToJest(pluginSettings)
   const configPath = pathToConfig(pluginSettings)
   const currentJestVersion = 20
-  const workspace = new ProjectWorkspace(pluginSettings.rootPath, jestPath, configPath, currentJestVersion)
+  const debugMode = pluginSettings.debugMode // TODO: add pluginSettings.debug
```
stephtr: Is that Todo still relevant?

src/messaging.ts
```diff
@@ -0,0 +1,55 @@
...
+import * as vscode from 'vscode'
+
+export interface MessageAction {
+  title: string
+  action: () => void
+}
+
+export function systemErrorMessage(message: string, ...actions: Array<MessageAction>) {
+  const msg = vscode.window.showErrorMessage(message, ..._extractActionTitles(actions))
+  if (msg) {
```
stephtr: In which case would `showErrorMessage` return something falsy? According to the documentation it returns `Thenable<string | undefined>` which, before being resolved, is always truthy.

src/messaging.ts
```diff
@@ -0,0 +1,55 @@
...
+  title: 'help',
+  action: () => vscode.commands.executeCommand('vscode.open', vscode.Uri.parse(TroubleShootingURL)),
+}
+
+export const TroubleShootingURL = 'https://github.com/jest-community/vscode-jest/blob/master/README.md#troubleshooting'
+//
+// internal methods
+//
+function _extractActionTitles(actions?: Array<MessageAction>): string[] {
+  if (!actions || actions.length === 0) {
```
stephtr: `map` also works with an empty array, the `!actions` check on its one should therefore be enough.

src/messaging.ts
```diff
@@ -0,0 +1,55 @@
...
+export function _handleMessageActions(actions?: Array<MessageAction>): (action?: string) => void {
+  return (action?: string) => {
+    if (!action) {
+      return
+    }
+    const found = actions.filter(a => a.title === action)
+    if (found.length === 1) {
+      found[0].action()
+    } else {
+      vscode.window.showWarningMessage(`unrecognized action ${action}`)
```
stephtr: Is there any other possibility for that to happen other than we messed up the actions parameter? Should we in that case really show a warning message?

src/messaging.ts
```diff
@@ -0,0 +1,55 @@
...
+export function systemWarningMessage(message: string, ...actions: Array<MessageAction>) {
+  const msg = vscode.window.showWarningMessage(message, ..._extractActionTitles(actions))
+  if (msg) {
+    msg.then(_handleMessageActions(actions))
+  }
+}
+
+// common actions
+export const showTroubleshootingAction: MessageAction = {
+  title: 'help',
```
stephtr: In my opinion it would look better to capitalize the button's first letter.

src/JestExt.ts
```diff
@@ -455,22 +503,6 @@ export class JestExt {
     this.jestProcessManager.stopAll()
   }
 
-  private getJestVersion(version: (v: number) => void) {
```
stephtr: Jest's --showConfig got only added in version 20.0.0, that's probably why we still relied on this `getJestVersion`. With your PR there would be no situation in which the "This extension relies on Jest 20+ features […]" warning gets displayed.
Since v20.0.0 is already older than one year, I wonder, if we could drop support for v19 (stopping to display the warning message)?

connectdotz:
@stephtr thanks for reviewing, good catch, I will address your inline comments in the next commit. 

> We should decide whether to drop the warning messages (and support) for Jest@19- or to (re-)add getJestVersion as a fallback version test.

First to clarify how the version info was used:
the version information is mainly used to help this plugin to parse jest output, it's obvious why we need that for jest test run, for settings (i.e. the run with --showConfig), we mainly wanted to grab the testMatch and testRegex patterns so we can check if a file is truly in the test run([here](https://github.com/jest-community/vscode-jest/blob/e39974085c3e0f98c87fd288c9ae451966a4d176/src/JestExt.ts#L364)), which triggers the editor visual display etc. vscode-jest doesn't use this version info other than showing misleading warning; and in jest-editor-support, it is also largely ignored because it is apparent if we can't parse the output and error will be reported back accordingly.

I wasn't aware `--showConfig` was added after 20, I agree in that case we don't need to consider the < 20 version check. However, the fallback is not getting back `getJestVersion()` because it won't give us testMatch and testRegex. Instead of considering how we can get the actual testMatch/testRegex, I wonder if we really need them to start with...? 

The more I think about it, the more I think we could just get rid of the whole Settings and checking-test-file logic... it is what jest do and we should just use its output and not duplicate its logic. Instead of checking the file patterns ourselves, why couldn't we just query the test results? But what if people didn't do the full test run, i.e. some tests are not in the test results? We will not have the full view of the test suites, which means some test files will have no (or unknown?) test indicator... and that will be correct, right? Just a thought, we might enoucter some other issues... granted it is a more radical change, but if it works, I think it will not only simplify the plugin, it will also make it more robust and correct... What do you guys think?

stephtr:
> The more I think about it, the more I think we could just get rid of the whole Settings and checking-test-file logic... it is what jest do and we should just use its output and not duplicate its logic. Instead of checking the file patterns ourselves, why couldn't we just query the test results?

I didn't think about that down to the last detail, but I think that would be very reasonable.

connectdotz:
@stephtr updates are in:
- addressed your review comments. 

- I was able to retire `jestSettings` but we still need to know if the file is "parsable", i.e. either typescript or javascript code. So replaced the test eligibility testing with simple file suffix check in `canUpdateActiveEditor()`

- Also discovered a new bug while working on this PR, see #349. Fixed it as well as it is really annoying. This also triggered some refactoring in `diagnostics.js` to allow diagnostics to be updated in 2 phrases: batch phase (existing) with only jest output, active phase (new) with actuate file content to correct line number issue. This should also improve batch phase performance.

- some refactoring and added more tests

I am not familiar with the snapshots, while I don't believe anything in this PR should impact it, the new jest version does bring in some snapshot changes. If you can test some snapshot use cases and windows that will be great.

Thanks.

connectdotz:
ping @stephtr @seanpoulter 

I asked @orta to add "review required" as the PR merge process prerequisite, hoping to make our PR process more robust. However, this does mean that reviewers will bear more responsibility in the merge process going forward. I made the assumption that this is something we, who loved vscode-jest enough to spend our free time improving it, are willing to take on... is that a wrong assumption?

If we can, let's try to wrap up this PR ASAP, as we are leaving our users hanging out there...

stephtr:
You are completely right and I'm sorry about that.
Unfortunately since a few weeks I'm that busy at work, that even my personal projects kind of stalled. Furthermore reviewing PRs is something I don't want to do in a hurry.
Concerning this PR: I looked at it and tested it on Windows, I'm fine with it.

connectdotz:
@stephtr I completely understand we all have lives outside of this project, thanks for taking the time to test and review 👍 while there are so many reasons not to...  

@Flavien-Pensato thanks for stepping up! 👍 We are all simple users, and you just proved that we can all contribute in so many different ways...  

the community never ceased to impress me, now let's finish the last mile... @orta do you think you can rebase/merge this PR, and if there is no other concern, maybe cut a release while you are at it?

mdurling:
Is a new version of this extension imminent? Our projects have been stuck on Jest 22 for a while.

orta:
Hah, doesn't look like it - definitely shipped now

screen shot 2018-07-23 at 3 25 44 pm

PR #356, opened 2018-07-28
Fix when "update snapshots" message is displayed #trivial

chrismcleod:
The existing check was not working in jest@23.4.1


chrismcleod:
I can definitely test the regex.  I took about 30 mins to try and figure out how to test that a notification is shown by a vscode extension.  I have never developed one myself however, and didn't get very far.  Any ideas or links with some information about how to test that?

chrismcleod:
@connectdotz I added at least one test for the error handler and pattern matching

Issue #351, opened 2018-07-20
Display in outline view


smith:
As of [VSCode 1.25](https://code.visualstudio.com/updates/v1_25#_document-symbols) the Outline view is included by default.

When editing Jest files, the `describe` and `it` blocks just show up as `<function>` everywhere.

It would be a lot nicer if these showed up as the string given to `describe`/`it` with an appropriate icon.

I *think* this could be implemented in vscode-jest using VSCode's DocumentSymbol API, but I haven't looked into it more than that.

smith:
In VSCode 1.30.0 this is done automatically. https://code.visualstudio.com/updates/v1_30#_languages

![image](https://user-images.githubusercontent.com/9912/49962831-f6ff0080-fedb-11e8-8c97-30e202cac316.png)

🔥

PR #357, opened 2018-08-06
Sidebar support

gstamac:
I implemented Test sidebar support for vscode. It is just a first version where you can see the status of test run and locate the test by clicking on it. It will also show the test result (with errors) in the tooltip. Next version should support some commands like run and debug. 
`jest-editor-support` has a shortcoming and will not parse describe blocks (described here #351). That means it cannot distinct between tests with the same name in the same file. So now when you click on such test it will always put the cursor on the first one. When describe block parsing will be implemented I can update sidebar to support it also.
Here are the screenshots....
image
image


gstamac:
Has anybody reviewed this?

connectdotz:
Usually, a PR is ready for review when it passes the ci tests, unless there is a good reason for an exception...

@gstamac looks like the ci test didn't finish travis test suite, maybe some new tests trigger async ops that stopped them from completing? Feel free to discuss here is you need some help.

gstamac:
It looks like Travis has problems with console.log. I removed it from a test (not my test) and now it works. I also added changelog entry.

src/SideBar/TestResultTree.ts
```diff
@@ -0,0 +1,71 @@
...
+  suites: TestResultSuite[]
+
+  constructor(results: JestFileResults, parsedResults: TestResult[]) {
+    this.name = results.name
+    this.suites = []
+    results.assertionResults.forEach(r => this.parseAssertionResults(r, parsedResults))
+  }
+
+  private parseAssertionResults(results: JestAssertionResults, parsedResults: TestResult[]) {
+    const suite = this.getSuite((<any>results).ancestorTitles, this.suites)
```
connectdotz: this crashed when the test file doesn't have the "describe" block outside of the "it" blocks, such as a default CRA app:

results.assertionResults=[{"failureMessages":[],"status":"passed","title":"renders without crashing"}]

> TypeError: Cannot read property 'forEach' of undefined
	at TestResultFile.getSuite (/Users/vsun/ConnectDotz/External/vscode-jest/out/src/SideBar/TestResultTree.js:28:16)
	at TestResultFile.parseAssertionResults (/Users/vsun/ConnectDotz/External/vscode-jest/out/src/SideBar/TestResultTree.js:11:28)
	at TestResultFile.results.assertionResults.forEach.r (/Users/vsun/ConnectDotz/External/vscode-jest/out/src/SideBar/TestResultTree.js:8:52)
	at Array.forEach (<anonymous>)
	at new TestResultFile (/Users/vsun/ConnectDotz/External/vscode-jest/out/src/SideBar/TestResultTree.js:8:34)
	at JestTreeProvider.loadTestResultsForFile (/Users/vsun/ConnectDotz/External/vscode-jest/out/src/SideBar/JestTreeProvider.js:35:16)
	at data.testResults.map.r (/Users/vsun/ConnectDotz/External/vscode-jest/out/src/SideBar/JestTreeProvider.js:30:58)
	at Array.map (<anonymous>)
	at JestTreeProvider.loadTestResults (/Users/vsun/ConnectDotz/External/vscode-jest/out/src/SideBar/JestTreeProvider.js:30:44)
	at JestTreeProvider.refresh (/Users/vsun/ConnectDotz/External/vscode-jest/out/src/SideBar/JestTreeProvider.js:15:14)
	at JestExt.updateWithData (/Users/vsun/ConnectDotz/External/vscode-jest/out/src/JestExt.js:304:30)

src/JestExt.ts
```diff
@@ -364,6 +369,8 @@ export class JestExt {
     const statusList = this.testResultProvider.updateTestResults(normalizedData)
     updateDiagnostics(statusList, this.failDiagnostics)
 
+    this.sidebarProvider.refresh(data)
```
connectdotz: do we need to process all the tests up front like this, giving we will only need to display the tree when the file is actually displayed in the editor... please see if you can move the logic to `triggerUpdateActiveEditor` instead.

connectdotz: indeed, I was thinking about the outline view, my bad.

trying it again on vscode-jest itself, added a few fake tests and encountered the following issues:
- looks like under the root "Tests" are the describe blocks instead of the file.  So if a test file doesn't have a root describe block or it has multiple parallel describe blocks, they will not show any hierarchical relationship. Is that intended? Is there a particular reason we don't want to use the test file for grouping?
- nested describe blocks are not shown.
- the root level `it` blocks are not shown, and if reload/restart vscode, it crashed the plugin, put it in the never-ending spinning "Starting watch mode" and the Test sidebar is empty... (same error as mentioned in earlier comment.)

gstamac: I added support for root tests, removed the folder name from filename and changed the default setting to show files by default. My practice has always been to have all tests inside describe and to have only one describe per file. That's why I chose the implementation I did.
I see you have to reload the window when the settings change but I think that's the case for all vscode-jest settings because `getExtensionSettings` is called on `extension.activate` event.

gstamac:
Finally had some time to look into this again. 

> if watch-mode only triggered partial tests, the test sidebar would show the partial results instead of the full test results like you intended.

I fixed the partial test run so it updates the results. The match is done based on file name so if you delete the test file partial run will not notice it. But if you stop/start jest it will rerun all tests and refresh the whole tree.

> I saw you now display the file nodes with the relative path, which is definitely shorter, but it could still truncate the actual file name. I wonder, instead of a flat file-path structure, how about a hierarchical one?

I don't agree with this. If you have a deep tree structure this will look terrible. I think this is OK for first version.

> can we have a command or shortcut to do the on-demand "deep" collapse or expand, it's like toggling the jest.autoExpandSidebar settings, so developers can quickly zoom in/out dynamically.

I believe this is more for vscode than vscode-jest. But later we can add some button to the top like mocha extension does. Again good enough for first version.

> in the explorer view, the file in the active editor gained autofocus; is it possible to do the same with the test sidebar? at least scroll/expand to the file level if not the test level by cursor...

You can go to the test or even line of the test failure already with a double-click.

gstamac:
Can somebody review this again. Thanks.

connectdotz:
@gstamac Thanks for following up , sorry for the delayed response, fell sick last 2 weeks...

I see the partial update is fixed, awesome! 👍 

>> can we have a command or shortcut to do the on-demand "deep" collapse or expand, it's like toggling the jest.autoExpandSidebar settings, so developers can quickly zoom in/out dynamically.

> I believe this is more for vscode than vscode-jest. But later we can add some button to the top like mocha extension does. Again good enough for first version.

that's fine.

>> I saw you now display the file nodes with the relative path, which is definitely shorter, but it could still truncate the actual file name. I wonder, instead of a flat file-path structure, how about a hierarchical one?

> I don't agree with this. If you have a deep tree structure this will look terrible. I think this is OK for first version.

A hierarchical tree structure is the standard way vscode display files (see Explore view), while UI is objective, I don't think vscode Explore look terrible, quite the contrary IMHO 😏. UI aside, the biggest problem I have with flat mode is that it hides the most important information - filename: 

sidebar-snapshot
(a snapshot from vscode-jest itself)

>> in the explorer view, the file in the active editor gained autofocus; is it possible to do the same with the test sidebar? at least scroll/expand to the file level if not the test level by cursor...

> You can go to the test or even line of the test failure already with a double-click.

I think you are thinking from the sidebar to file, but I meant the other direction. From the test file, I would like to see the sidebar scroll to the active file accordingly. 

---

please also take a look of the comments earlier regarding setting changes not get reflected, you would need to add the new settings in `triggerUpdateSettings`.

I will take a more indepth look at the code once we settle the discussion above. Thanks for your patient.

gstamac:
I've implemented settings update on `triggerUpdateSettings`.

> A hierarchical tree structure is the standard way vscode display files (see Explore view), while UI is objective, I don't think vscode Explore look terrible, quite the contrary IMHO 😏. UI aside, the biggest problem I have with flat mode is that it hides the most important information - filename:

Regarding file structure I understand your issues but I believe it shouldn't be used anyway. For tests there is another tree structure defined with `describe` blocks and I believe this should be the tree displayed in the sidebar. If you look at Mocha sidebar it doesn't have a file tree structure. I implemented it per your request but I still think it shouldn't be used. If somebody has a need for it and wants to add features to it they can do that.

> I think you are thinking from the sidebar to file, but I meant the other direction. From the test file, I would like to see the sidebar scroll to the active file accordingly.

I think this is also related to the last request you had. I would see a better feature to scroll to the test you're on but I'm afraid this is too difficult to do with the current API. It would be too time intensive and it would make editing impossible. There is a feature request for VSCode (#5455) which would make it possible to click on the gutter decorator and go to the test. I believe this would be a much better solution.

connectdotz:
> Regarding file structure I understand your issues but I believe it shouldn't be used anyway. For tests there is another tree structure defined with describe blocks and I believe this should be the tree displayed in the sidebar.

As a tool provider, we don't get to decide users' test structure or practice. You might have a strong preference and good reason for having a root `describe` block, it doesn't mean other users' preferences or reasons are less valid. Actually, if you look at a few popular open source projects that use jest, you will see many of their tests do not have root describes: [jest](https://github.com/facebook/jest/blob/master/packages/jest-cli/src/__tests__/notify_reporter.test.js), [yarn](https://github.com/yarnpkg/yarn/blob/master/__tests__/integration.js), or the default react apps generated by CRA. For projects without root describe blocks, the `test file` is its scope, equivalent to the root describe, therefore should be displayed accordingly; and the most intuitive UI is to adopt vscode's standard hierarchical tree struture that the Explore view used. 

A good tool adapts to users environment, not the other way around. I am not arguing which test structure is better, only to emphasize that, as a tool provider, we need to support both. Having a setting `showFilesInSidebar` as you implemented is a great way to do this: for projects using root describes, they could simply turn off `showFilesInSidebar` and you got mocha like display; for the others, turn on `showFilesInSidebar`. Now, we just need to make sure the files do display clearly and intuitively when the flag is on.

---

> I would see a better feature to scroll to the test you're on but I'm afraid this is too difficult to do with the current API...There is a feature request for VSCode (#5455) which would make it possible to click on the gutter decorator and go to the test.

indeed it would be nice if we can scroll to the particular test when clicked, but looks like this request has been sitting for over 2 years, i.e. unlikely to be implemented soon, I think the next best thing is to scoll and expand the file/root-describe when active doc changed, once the test level event is available, it can be easily enhanced. 

JextExt already have a `triggerUpdateActiveEditor` to handle active editor change event, you can simply add the  following tree view code there.  Reading up the vscode [api](https://code.visualstudio.com/docs/extensionAPI/vscode-api#_workspace), it seems that you can get a TreeView by using `createTreeView` instead of `registerTreeDataProvider`; once you have the [TreeView](https://code.visualstudio.com/docs/extensionAPI/vscode-api#TreeView) instance, you can scroll to the position with `reval`; and of course you already know how to do item expansion. So it seems relative simple, but I haven't implemented it myself so could be wrong...

smith:
@gstamac let me know if there's anything I can do to help get this finished up. Thanks for adding it!

gstamac:
@smith Sorry haven't had time to work on this for a while. I renamed the config names. File display issue is not important in my opinion because that's not how it should be done anyway. If you look at Mocha addon it doesn't even support filename display. If you believe it's not done correctly I can remove it.

gstamac:
@smith Have you had time to check this?

smith:
@gstamac no it's kind of fallen of my radar for now. Sorry.

connectdotz:
@gstamac we are preparing for the next major release that will incorporate multi-root and hopefully sidebar support too. see #442 

I was trying to merge this PR to `beta` but multi-root has introduced some structural change that I figure you might want to do the merge yourself. It would be great if you can rebase this PR against `beta` branch, then I will merge it and help to get this feature out for the next release.

gstamac:
@connectdotz I'll try to do it this weekend.

escaton:
@gstamac do you need any help with rebase?

rhalaly:
The images show that the test result indicator implemented as a circle with color indicator: green (pass) / red (failed). 
But in [this PR](https://github.com/jest-community/vscode-jest/pull/512), they changed the indicator to an icon (✘, ✔) for colorblind people. 

I think that this feature should also use the icon indicator for colorblind people and for uniform look & feel...

gerich-home:
Hey guys, I see last comment in this PR on December. Any chance I could help with finishing this PR?
What should be done? (any checklist?)

connectdotz:
@gerich-home If I remember correctly, it was last discussed that we should use vscode's standard hierarchical tree structure (like the Explore view)  when `showFilesInSidebar` is true`. Would be great if we can get this completed 👍

wmertens:
The sidebar view is super nice, it gives way clearer information than the console window. Would it not be an option to merge this as-is, and add the file tree later?
There are some other sidebar extensions, but they don't have all the other integrations that vscode-jest has.

rossknudsen:
Just as a heads up - I've raised #625 which makes the test results from this extension available to other extensions.  The aim being to enable an extension that I maintain to have access to the test results so they can be displayed in a side bar using [Vs Code Test Explorer](https://marketplace.visualstudio.com/items?itemName=hbenl.vscode-test-explorer).  It might be desirable to not merge this and not reinvent the wheel.

Either way, feel free to comment on #625 as I'm pretty keen to get it across the line.  I've done some initial spike work with my extension and it seems to be able to access `vscode-jest` and get updates when the test results change.

connectdotz:
close this in favor of the upcoming vscode testing API/Explorer. @gstamac sorry we did not reach the action plan sooner but thanks for the contribution nevertheless.

PR #358, opened 2018-08-07
Update create-react-app up to date and add vscode setting to enable vscode-jest

tristola:
- Updated create-react-app
- I added .vscode setting file to enable vscode-jest

Probably users will prefer to put is as default in their own user settings,
but this makes testing the plugin with create-react-app easier.


tristola:
Agreed. Moved it to a separate config markdown file.

PR #361, opened 2018-08-15
Fix mapping from Native Windows to WSL Paths

mojadev:
This is a naive fix that allows jest runs in WSL (Windows Subsystem for Linux)
to be executed and also fixes mapping wsl paths to the windows paths (which
are required by the vscode instance that is not running in WSL). Parts
of this fix should be moved to jest-editor-support.

The idea was to touch as little logic as possible and only introduce wsl specific logic in the spawning of the process and the parsing of resultpaths.

- JestProcess (I have no clue why the diff says everything has changed): The 
```
    // If Windows Subsystem for Linux is used, spawn in wsl
    if (this.projectWorkspace.pathToJest && this.projectWorkspace.pathToJest.startsWith('wsl')) {
      options.createProcess = createProcessInWSL
    }
```
snippet is a really dirty workaround which I would try to circumvent by updating the jest-editor-support in the aftermath. Having a useWSL Flag would really make sense in the ProjectWorkspace, even if that means it becomes an even bigger data structure

- WslProcess would also end up in jest-editor-support, or the process implementation there. 

- TestResult has been updated to rewrite the WSL (/mnt/c) paths to Windows paths (C:\)

### DIFF
--- a/src/JestProcessManagement/JestProcess.ts
+++ b/src/JestProcessManagement/JestProcess.ts
@@ -1,6 +1,7 @@
 import { platform } from 'os'
-import { Runner, ProjectWorkspace } from 'jest-editor-support'
+import { Runner, ProjectWorkspace, Options } from 'jest-editor-support'
 import { WatchMode } from '../Jest'
+import { createProcessInWSL } from './WslProcess'
 
 export class JestProcess {
   static readonly keepAliveLimit = 5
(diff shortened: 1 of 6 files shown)


mojadev:
Thanks for looking over it - I will move parts to jest-editor-support as soon as I find some minutes - then I can also use a dedicated flag as I can pass it down using the project workspace, hence getting rid of this ugly wsl detection hack.

Regarding wslpath: I didn't use it as I expect it hit performance rather hard if I do it for every path, but I think it would be sufficient to only convert the projects base path once and then use that one - I'll give that a try. Right now it feels there's a lot of guesswork involved in the path handling and I too would like to get rid of that..

mojadev:
To give a short update here: I moved all the wsl path resolution out to a own package https://github.com/mojadev/wsl-path) and hope to find some tme this evening to wrap up the rest of the PR, as now it's only about detecting wsl (I would use an own setting useWsl, as this already exists in launch configurations of vscode) and calling the path mapping before calling wsl and when parsing he results

PR #364, opened 2018-08-21
Make debug button respect custom jest config location

ThomasRooney:
Fixes 'Debug' button ignoring configuration option 'pathToJest'.

Can be replicated by breaking a test on core jest repo and clicking 'Debug' option -- it is shown that the jest config defined in .vscode/settings.json is ignored.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ Bug-fixes within the same version aren't needed
 -->
 
 ## Master
-* Fixes 'Debug' button ignoring configuration option 'pathToJest'.
+* Fixes 'Debug' button ignoring configuration option 'pathToConfig'.
 
 ### 2.9.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Bug-fixes within the same version aren't needed
 
 -->
 
+## Master
+* Fixes 'Debug' button ignoring configuration option 'pathToJest'.
+
 ### 2.9.0
 
 * Adds a setting to control when the debug CodeLens appears - seanpoulter
--- a/src/DebugConfigurationProvider.ts
+++ b/src/DebugConfigurationProvider.ts
@@ -1,17 +1,20 @@
 import * as vscode from 'vscode'
+import { ProjectWorkspace } from 'jest-editor-support'
 import { getTestCommand, isCreateReactAppTestCommand } from './helpers'
 
 export class DebugConfigurationProvider implements vscode.DebugConfigurationProvider {
   private fileNameToRun: string = ''
   private testToRun: string = ''
+  private workspace: ProjectWorkspace
 
   /**
    * Prepares injecting the name of the test, which has to be debugged, into the `DebugConfiguration`,
    * This function has to be called before `vscode.debug.startDebugging`.
    */
-  public prepareTestRun(fileNameToRun: string, testToRun: string) {
+  public prepareTestRun(fileNameToRun: string, testToRun: string, workspace: ProjectWorkspace) {
     this.fileNameToRun = fileNameToRun
     this.testToRun = testToRun
+    this.workspace = workspace
   }
 
   resolveDebugConfiguration(
@@ -71,6 +74,9 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
       debugConfiguration.protocol = 'inspector'
     } else {
       // Plain jest setup
+      if (this.workspace.pathToConfig) {
+        debugConfiguration.args = [...debugConfiguration.args, '--config', this.workspace.pathToConfig]
+      }
       debugConfiguration.program = '${workspaceFolder}/node_modules/jest/bin/jest'
     }
 
--- a/src/DebugConfigurationProvider.ts
+++ b/src/DebugConfigurationProvider.ts
@@ -5,7 +5,7 @@ import { getTestCommand, isCreateReactAppTestCommand } from './helpers'
 export class DebugConfigurationProvider implements vscode.DebugConfigurationProvider {
   private fileNameToRun: string = ''
   private testToRun: string = ''
-  private workspace: ProjectWorkspace
+  private workspace: ProjectWorkspace = {} as ProjectWorkspace
 
   /**
    * Prepares injecting the name of the test, which has to be debugged, into the `DebugConfiguration`,
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -402,7 +402,7 @@ export class JestExt {
     const restart = this.jestProcessManager.numberOfProcesses > 0
     this.jestProcessManager.stopAll()
 
-    this.debugConfigurationProvider.prepareTestRun(fileName, identifier)
+    this.debugConfigurationProvider.prepareTestRun(fileName, identifier, this.workspace)
 
     const handle = vscode.debug.onDidTerminateDebugSession(_ => {
       handle.dispose()
--- a/tests/DebugConfigurationProvider.test.ts
+++ b/tests/DebugConfigurationProvider.test.ts
@@ -2,6 +2,7 @@ jest.unmock('../src/DebugConfigurationProvider')
 
 import { DebugConfigurationProvider } from '../src/DebugConfigurationProvider'
 import { getTestCommand, isCreateReactAppTestCommand } from '../src/helpers'
+import { ProjectWorkspace } from 'jest-editor-support'
 
 describe('DebugConfigurationProvider', () => {
   it('should by default return a DebugConfiguration for Jest', () => {
@@ -33,7 +34,23 @@ describe('DebugConfigurationProvider', () => {
     expect(config.args).toContain('--env=jsdom')
     expect(config.args).toContain('--runInBand')
   })
+  it('should utilise workspace jestConfig', () => {
+    const folder = { uri: { fspath: 'folderName' } }
+    const fileName = 'fileName'
+    const testNamePattern = 'testNamePattern'
+    const workspace = {
+      pathToConfig: './jest-config.js',
+    }
+    const expected = ['--runInBand', '--config', workspace.pathToConfig]
 
+    const sut = new DebugConfigurationProvider()
+    sut.prepareTestRun(fileName, testNamePattern, workspace as any)
+
+    const configuration = sut.provideDebugConfigurations(folder as any)
+
+    expect(configuration).toBeDefined()
+    expect(configuration[0].args).toEqual(expected)
+  })
   it('should append the specified tests', () => {
     const fileName = 'fileName'
     const testNamePattern = 'testNamePattern'
@@ -41,7 +58,7 @@ describe('DebugConfigurationProvider', () => {
     let configuration: any = { name: 'vscode-jest-tests' }
 
     const sut = new DebugConfigurationProvider()
-    sut.prepareTestRun(fileName, testNamePattern)
+    sut.prepareTestRun(fileName, testNamePattern, {} as ProjectWorkspace)
 
     configuration = sut.resolveDebugConfiguration(undefined, configuration)
 
--- a/tests/JestExt.test.ts
+++ b/tests/JestExt.test.ts
@@ -146,7 +146,7 @@ describe('JestExt', () => {
       expect(configuration).toBeDefined()
       expect(configuration.type).toBe('dummyconfig')
 
-      expect(sut.debugConfigurationProvider.prepareTestRun).toBeCalledWith(fileName, testNamePattern)
+      expect(sut.debugConfigurationProvider.prepareTestRun).toBeCalledWith(fileName, testNamePattern, projectWorkspace)
     })
   })
 


stephtr:
When revamping the debug feature, we [intentionally decided against](https://github.com/jest-community/vscode-jest/pull/287#issuecomment-378718821) including `pathToJest` and `pathToConfig`. I'm still a bit unsure about that, but let's see what @connectdotz currently thinks about it.

ThomasRooney:
Ah -- I didn't know this was an intentional omission. It seems an odd behaviour because I expected that when a test fails, the selecting of the debug button would run the tests under the original conditions that the test failed in.

In particular, the codebase I work with names jest tests *.jest.js. Since the test file pattern ignored these, the debug button was not finding and running the associated test; I assumed this was a bug.

I am of the opinion that this behaviour should be changed, but I may be missing context on how other people use the feature.

connectdotz:
> I expected that when a test fails, the selecting of the debug button would run the tests under the original conditions that the test failed in.

Not necessary... if you look at the debug config, you will see they are not quite the same as the plain jest run, especially when you consider react/react-native and other frameworks. Also sometimes we might need different behavior such as coverage for regular but no for debug...

While incorporating `jest. pathToConfig` in default debug config seems to be a reasonable idea, how about `jest.pathToJest`? Wouldn't that be inconsistent to just use one but ignore the other? And as @stephtr mentioned in the original PR, to include `jest.pathToJest` is, well, complicated and error prone... 

(A side point here, I was always wondering why having a separate `jest.pathToConfig` setting instead of just appending the config as part of npm script or `jest.pathToJest`? )

Anyway, instead of implementing another set of elaborated guesses (and still be wrong) for the debug config, IMHO, it will be cleaner and definitely simpler to just provide template based config like @stephtr did. I did a quick issue search to see if there is any complaint about debug config and `jest.pathToConfig` but found none. So maybe for most users who are not as entrenched in vscode-jest as we are, adding customization in vscode's debug config is not that outrageous... 

@ThomasRooney is your issue resolvable with a custom debug config?

ThomasRooney:
>> @connectdotz using a custom debug config is a workaround, but to my understanding the custom debug config is not connected to the 'Debug' button in the UI. If these were connected (i.e. something like the pressing of the 'debug' button invoked the launch.json command with name 'Debug Tests' and an additional argument to precisely specify the test that was desired to run) then this issue might be considered resolved, because a user has a capability to make the 'Debug' button work on their codebase.

Actually, this looks like I misread the documentation. Using name `'vscode-jest-tests'` utilises that template for running the given tests; and so this my issue can be resolved that way. 

Default behaviour might still be worth changing though; I still consider this unintuitive.

Updating the vscode-jest repository such that the 'debug' button works via renaming its debug configuration to 'vscode-jest-tests' would be a good step in improving the documentation; I looked to this repositories configuration to try and resolve my issue as an example and jumped to 'bug' when it also didn't work.

geekox86:
Greetings,

Thank you for you great plugin.

Why the debug facility respects the custom Jest configs only if defined inside package.json?

When I defined my Jest configs in jest.json or jest.config.json next to my package.json with ot without jest.pathToConfig in VS Code settings.json, it is not respected.

I really think this complicates the debug configuration when using TypeScript source and test files, as I will need to define a custom debug luncher in VS Code lunch.json and elaborate debug script.

connectdotz:
@geekox86 according to [jest document](https://jestjs.io/docs/en/configuration.html):

> Jest's configuration can be defined in the package.json file of your project, or through a jest.config.js file or through the --config <path/to/js|json> option

please try `jest.config.js` if you haven't... jest should be able to automatically pick up this file without any customization.

geekox86:
Thank you for your reply.

The plugin does not detect jest.config.json while debugging.

Any clue?

connectdotz:
it should. if you run jest on the terminal and it also doesn't detect the `jest.config.js`, then something wrong with the way you set up the config. If you can't figure it out, please file a separate issue with more detail such as the error message shown on your terminal, version and sample repo etc...

nelson-perez:
Good I was working on a fix last Friday for this so this is great.

stephtr:
> @stephtr I don't understand this comment; given that current behaviour is just running the hardcoded '${workspaceFolder}/node_modules/jest/bin/jest' it seems reasonable behaviour to instead run jest.pathToJest if its defined? I don't see why it would be complicated and error prone, surely replacing debugConfiguration.program = '${workspaceFolder}/node_modules/jest/bin/jest' with debugConfiguration.program = this.workspace.pathToJest (when pathToJest is defined) is enough?

That would work for directly running jest, but not, when `pathToJest` is pointing to a script, which spawns a new process for running jest (like create-react-app is doing). In that case one would have to use the [`runtimeExecutable` instead](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#debugging-tests-in-visual-studio-code) of the `program` attribute (even if setting `"autoAttachChildProcesses": true` ). My problem was, that I found no reliable way to automatically decide between those two properties.

> _Default behaviour might still be worth changing though; I still consider this unintuitive._
> noted. Sometimes the intuitions don't converge or a solution that would invite more problems than it solved... unfortunately, at such crossroads, we will have to make tough trade-off decisions. But we should definitely keep our eyes open for a better solution when available...

What would be the downside of changing the behaviour to automatically pick up `pathToJest` and `pathToConfig`? Is using a custom Jest path/config for watching tests but the default path/config for debugging tests that common?

nelson-perez:
Our team is in the same boat we do have a custom path for the Jest configuration and can’t really use it if the extension does not respect the jest.pathToConfig. 

I do like jest but the lack of tooling for us does make it hard to use without a good workflow and this extension is the only one that gets us closer to some tooling for mocha of debugging individual tests. 

Also I think is unintuitive that by default to NOT show the Debug button when the test passes at first it felt like a bug then once you realize the behavior need to add an assert.fail to just debug an issue that is not caugh by the test.

Then reading the documentation I figured that I could configure it to show in any state but why should I need to do that by default. It feels unreasonable the need to add a .vscode/setting.json in every project just to modify the behavior of an extension.

Just remember that we write software so it can do the work for us and not for us to have to work for the software.

seanpoulter:
Sorry for coming out of nowhere on this one.

> @nelson-perez :
> can’t really use it if the extension does not respect the jest.pathToConfig.

That's a fair expectation and I'm disappointed if we go the other direction. It'd break the use case for half my team if we were using Jest.
<br/>

> @nelson-perez :
> I think is unintuitive that by default to NOT show the Debug button when the test passes

Nah, I chose those [those default settings](https://github.com/jest-community/vscode-jest/blame/master/package.json#L103). When a test is passing I don't want to debug it and am unlikely to revisit it. If it's a problem, please create an issue to discuss the defaults off-thread.
<br/>

> @nelson-perez :
> It feels unreasonable the need to add a .vscode/setting.json in every project just to modify the behavior of an extension.

If you prefer the Debug CodeLens everywhere, I'd suggest you add a user level setting instead of per workspace. Refer to these docs https://code.visualstudio.com/docs/getstarted/settings and feel free to open an issue if there's a problem merging the user and workspace level settings. I haven't tried it.

nelson-perez:
@seanpoulter I don't see why respecting the jest.pathToConfig while debugging which is exactly what the extension does on the background (every time you save a file) would brake the use cases for your team.

On my organization I'm the one that sets the best practices, code patterns and developer workflows for our entire organization to make everyone more productive and their life's easier. So it does not make sense to ask each individual on the entire team to create an user specific settings in order to have a consistent developer workflow experience on our entire organization.

I'll create an issue show the codelens for debugging even when the test is passing because I don't see why it should not be there as It is not inconvenient to see the codelens if the test passes but it is an inconvenient to not see the codelens when you want to debug a test that already passed like any other tool that allows you run specific tests.

stephtr:
> I'm trying to agree with you that we should aim to support the config when we launch the debugger.

👍

> It's not there because skipping over that CodeLens for every passing test adds a tax that I don't want in my workflow.

What would you think of hiding the debug functionality for passing tests behind a quick fix (lightbulb)? (One could then maybe also remove the `debugCodeLens.showWhenTestStateIn` setting.)

> Should we fire up a Google Slides doc to collaborate on that?

I'm not sure if we even need that. In my opinion it's just:
* try to run the `vscode-jest-tests` config from `launch.json`
* if that fails: check if the project has been bootstrapped by create-react-app
* if not: use `pathToJest` and `pathToConfig` for building the DebugConfiguration

It won't work perfectly (since we can't predict whether the executed command will run Jest in the existing process or create a new node process), but in those cases one can still switch to a custom debug configuration.

connectdotz:
Looks like I am the minority here... ;-) 

- "respect config location at debug time", what's wrong with that? 
- "helping us to use jest in vscode is what this plugin do"
- "software is used to help us not the other way around"

All of these statements are correct, but consider:
- we don't control vscode debugger as we can with jest process. The way to debug in vscode is via launch.json, not settings.json. 
- every initial request is all very reasonable, otherwise we won't be discussing it... the question is not "if" we should help, but "how".  For some users, maybe adding `jest.pathToConfig` to debug config is all they needed, but for others, maybe not. We can keep enhancing our logic to generate a default debug config so users don't have to, or we can cut through all these guesses and focus on helping users to set up their own debug config. 
- Why do I favor later? because we just lived through an example: when users complaint the plugin didn't work because it couldn't find `jest`. Instead of educating people to use `jest.pathToJest`, we started to add logic to craw node_modules, yes it worked for some but not for monorepo, react, and npm scripts, to name a few; so we started to parse package.json, check react scripts...; and if react has different patterns for different versions? let's parse its package.json too!... you can see there is nothing we **"can not"** do, the question is always _"is this the best solution for the problem?"_. The last thing we want is to turn the solution into a new problem... eventually, the code became so complicated that it started to break and caused many grief and sad users... Just imagine if we had told our users on day 1, all they needed to do is to customize `jest.pathToJest`, it would take 10 seconds and everybody can do it without waiting for the plugin code change... 

Instead of adding logic to generate a hidden default debug config, personally, I think I would rather spend time in doing this: when user click on the debug button and there is no existing jest debug config, we will walk them through the setup process to create their jest debug config. Then there will be no expectation misalignment - "if you give me a debug button, then when I click on it, it should just work" - and it will work for all users, all environments...

As I mentioned earlier, sometimes our opinions won't converge, and that's all right, I could be wrong anyway. If @stephtr or @seanpoulter would sign off on this PR, I can certainly understand and won't hold you guys up.

stephtr:
> Instead of adding logic to generate a hidden default debug config, personally, I think I would rather spend time in doing this: when user click on the debug button and there is no existing jest debug config, we will walk them through the setup process to create their jest debug config.

I would also be perfectly fine with that. However a larger obstacle for me was that VS Code doesn't offer an API for programmatically adding a debug configuration to the `launch.json` file. The team behind the OmniSharp extension, having the same issue, therefore [manually extends and writes](https://github.com/OmniSharp/omnisharp-vscode/blob/2330359e355b14451411b2353ff3acf7630044fa/src/assets.ts#L482) the debug configuration file, which would be also possible for us.

Would displaying a (smartly prefilled) dialog when clicking Debug for the first time be a feasible way for everyone?
(A la "Which debug preset do you want to use? [Pure Jest setup] [Create-React-App] [Custom]" "Which command has to be run for testing? [default: command from `pathToJest`]" "Should any additional parameters be included? [default: external config from `pathToConfig`]")

connectdotz:
> Would displaying a (smartly prefilled) dialog when clicking Debug for the first time be a feasible way for everyone?

I think it is a good start to have the dialog walk through the setup workflow using the action buttons. At the end of the workflow, if we can write out the config "template" that will be great, otherwise I think at the minimal we could do: 
1. enhance our debug document. It probably should be a separate file so we can add a workflow and step-by-step instructions accordingly. Basically, we should be able to articulate everything we want to do in the dialog system, which really is just an implementation of such process.
1. Once we have such doc, even if we can't write out the debug config, we can at least point users to a specific section of the doc so they can carry out the rest.

It is important that we stress this config is just a "template", i.e. a starting point that users can, and should, adjust to fit their environments. We can provide a few examples and common scenarios in the document, such as adding custom config, jest location, reuse existing debug config... but **ultimately, users are in full control of their own debug config**. BTW, This config (launch.json) can certainly be checked into git so the whole team can share.

food for thought: we could use this kind of worflow dialog, at first time launch, to  setup the `jest.xxx` variables, another sore spot for this plugin, and possible debug config template at the same time => usability + transparency + efficiency...

tostringtheory:
Just realized this PR existed; been talking about this 'issue' in #409. 

I also faced this friction when trying to debug in vscode and it wasn't respecting the `pathToConfig`, which I would've figured it should use that since I DID set the setting.  Looks like the PR's stagnated.  Should this just be closed/abandoned?  If it'd be closed, I think there should at least be some clarification about this in the readme - "even if you set a custom configuration path, that is not used for the debug command.  We treat debugging as a separate concern than watching your tests.  To customize the debug command functionality, [...]".  As I mentioned in #409 , the small blurb about the custom launch.json command wasn't originally clear to me in intent.

seanpoulter:
> Looks like the PR's stagnated. Should this just be closed/abandoned?

Given that the reviewers, including myself, haven't moved this ahead I'd like to apologize to @ThomasRooney and shift the conversation on to #409. @connectdotz and @tostringtheory have started to have valuable conversations about how we'll handle the debugging. Please join them in that conversation and let's drive this forwards. If you're still interested in the extension, I'd encourage everyone to participate in the design discussions, development, and code review.

Issue #342, opened 2018-07-03
vscode-jest does not start when the workspace folder name (multi-root) contains blanks


marcelh-gh:
### Environment

1. `node -v`: 10.5
2. `npm -v`: 6.1
3. VSCode: 1.24.1
4. Operating system: Windows 10 1709

connectdotz:
sorry, it is not clear what "blanks" mean? blank space of folder name? empty folder? does it work when the folder is not "blank"? best to provide an example repo as the template has suggested...

marcelh-gh:
blank spaces... like "VS Code"... as soon as they are replaced, everything is fine.

connectdotz:
interesting... vscode-jest doesn't officially support multi-root, i.e. it only operates under the root... is this folder the root folder or a sub-folder?

marcelh-gh:
@connectdotz thanks a lot for the quick replies! :) the main folder for the multi-root workspace had blanks in its name. under that folder there are root-folders for each sub-project.

guylando:
@connectdotz Any workaround? I also cant use jest vscode plugin because my project has a parent folder which contains a space inside its name.
Maybe somehow using windows cmd "subst" command to map the folder to "P:\" and run from there?

connectdotz:
this is indeed a bug that we did the plain string split when spawning the process in `Process.js` of `jest-editor-supports` package (this fix should be pretty straight forward, if anybody is interested in bmitting a PR to address this, I can help answering any question you have...)

 meanwhile you can try to use the scripts in package.json to hide the path then customize your `js.pathToJest` to point to it. For example:

in your root package.json, add a test script if not already exist
```
"scripts": {
   "test": "jest" // make sure you can run this script from commandline
}
```
in .vscode/settings.json, customize `jest.pathToJest`:
```
"jest.pathToJest": "yarn test" // or `npm run test --`
```
that should get you around the bug until it is fixed.

guylando:
@connectdotz this workaround doesnt seem to work if jest is not installed globally. And if I pass to npm run the full path of jest script then there is an error "script not found"

guylando:
@connectdotz I get now:
ERR! code ELIFECYCLE
and from some reason the plugin tries to run:
 ERR! Project@1.0.0 test: `jest --no-cache "C:\Users\user1\AppData\Local\Temp/jest_runner.json"`

why does it add /jest_runner.json with a forward slash?

guylando:
Fixed by using:
[...]

Issue #254, opened 2018-02-08
Multi-Root Workspace


jpokrzyk:
Does this extension support a VS Code multi-root workspace?

Couldn't seem to find any existing questions relating to a multi-root workspace. When I load a multi-root workspace only the tests from the first project of the workspace are run.

seanpoulter:
Sounds related to #129?

jpokrzyk:
Read through that. Related but slightly different. I don't have a top-level `package.json` or even a top level folder I can use for an override file.

For now, I just re-ordered the folders node in the workspace config so the project I want to run is the first one on the list.

I just started using this extension about an hour ago and I'm working on some other issues. I'll take a crack at a PR for multi-root. Just wanted to make sure I wasn't missing something.

I'll let everyone know here when/if I start working it so people aren't waiting for a PR that hasn't been started.

orta:
Good luck yeah, I don't think this supports multi-root at all

rjokelai:
Vscode has a pretty good description on how to modify plugins to work for multi-root workspaces:
https://github.com/Microsoft/vscode/wiki/Extension-Authoring:-Adopting-Multi-Root-Workspace-APIs

connectdotz:
hi, multi-root support has finally arrived, feel free to give it a try: [3.0.0-preRelease](https://github.com/jest-community/vscode-jest/releases/tag/3.0.0)

Issue #370, opened 2018-08-27
Coverage overlay does not appear if coverage was collected with `jest --coverage` rather than `collectCoverage: true` in `jest.config.js`


garyking:
### Environment

1. `node -v`: v8.11.2
1. `npm -v`: 6.4.0
1. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): jest@23.5.0
1. your vscode-jest settings if customized:

```
  "jest.enableCodeLens": false,
  "jest.pathToJest": "node_modules/.bin/jest",
```

1. Operating system: Mac 10.13.5 (17F77)

### Prerequisite
- are you able to run jest test from command line? Yes
- how do yo run your tests from command line? (for example: `npm run test` or `node_modules/.bin/jest`) `jest --coverage`

### Steps to Reproduce

If a project does not have `collectCoverage: true` in `jest.config.js`, and instead collects coverage through `jest --coverage`, then this extension does not show the coverage overlay for the project. I'm talking about the small text at the top of the file does not appear, and the color highlighting does not appear, when using Toggle Coverage Overlay, even though there is uncovered code in the file.

The output of both methods of coverage collection result in the exact same HTML output, so I don't know why it would matter, but it definitely seems to. I'm guessing because this extension only looks for that key in the config, rather than checking the existence of the coverage files themselves?

If I add `collectCoverage: true` to `jest.config.js`, then run `npx jest`, then reload VSC, then the overlay will work.

connectdotz:
what if you do this:
```
"jest.pathToJest": "node_modules/.bin/jest --coverage"
```

crichmond1989:
I have the same issue. I tried using "node_modules/.bin/jest --coverage", but that didn't help. Also gave "npm run cover --" a try. My cover script is: "jest --coverage".

rschoch:
I can attest to the same issue and i don't want to put the collectCoverage: true in the jest.config.js because it obviously is slower in test execution time. any workarounds lads?

connectdotz:
can you guys provide a sample repo?

nasreddineskandrani:
same issue. I can't see the overlay.
Can i use `collectCoverage: true` in package.json jest config. Seems not working?

nasreddineskandrani:
@connectdotz i'll provive a sample repo before the end of the week.

nasreddineskandrani:
[...]

PR #379, opened 2018-09-16
Multi root support

escaton:
Address #254, partially #342

- [x] instantiate jest per workspace folder
- [x] prompt for workspace when calling commands
- [x] share codeLens providers across all jest instances
- [x] separate resource and window scoped settings according to [adoption guide](https://github.com/Microsoft/vscode/wiki/Extension-Authoring:-Adopting-Multi-Root-Workspace-APIs#settings)
- [x] fix tests
- [x] add new tests for `extension.ts`
- [ ] ~~gracefully handle incompatible workspaces (only some of them could trigger activation)~~
- [x] improve statusbar (prioritize messages from different extension instances)
- [x] fix multiple TestsResults displayed in first opened file
- [x] detach DiagnosticCollection from JestExt
- [x] add `(enabled,disabled)WorkspaceFolders` settings to manage jest running in multiroot environment


escaton:
Need advice. How to handle incompatible workspace correctly? Previously any crash raised an exception. And that was reasonable, because in most cases extension was activated in proper workspace due to "activationEvents" list. But now only one of several workspaces could be compatible. How it should be handled?

connectdotz:
@escaton sorry for the late response.

> Previously any crash raised an exception. 

did this plugin or vscode throw exception? when and where? can you be more specific of the code or maybe stacktrace?

Without knowing too much detail and constraint, on the principle, it seems like a reasonable approach for the mutliroot project to ignore an "incompatible workspace" (I assume you meant a workspace not running jest?), i.e. fail silently.

o2genum:
I don't know much about how the extension works internally, but this functionality + ignoring incompatible workspaces would be a good start. Super useful for monorepos.

escaton:
Turns out its not that easy to mute exceptions specifically for that case when jest failed to start in incompatible folder.
Instead i just add two window scoped settings (`enabledWorkspaceFolders, disabledWorkspaceFolders`) to decide whether to start jest or not.
If both contain same item — don't start (disabling priority), if both doesn't contain — start.

So far looks like this is done. Please give it a try and share feedback :)

o2genum:
Tried it. It works, I added the root workspace without jest to `disabledWorkspaceFolders`, and it worked too.

Can't get coverage overlay to work, and items from Problems list disappear after I click on them, but these must be separate issues.

escaton:
Thank you, i reproduce coverage problem on latest version from store, can you check it too?

o2genum:
Yep, the latest version from store doesn't show coverage.

interrobrian:
Has there been any progress on this? My team is switching to multi-root for various reasons, and we're unable to use this extension until this is resolved.

LennyPenny:
@escaton any progress on the coverage issue?

escaton:
@LennyPenny looks like there is separate issue about coverage #370 
Could you check, if you have coverage with current version of extension from store?

escaton:
@connectdotz can i ask your help for validating coverage problem?

lvl99:
If the coverage is an existing issue, then that could be solved with a different PR.

I reaaaaaaaaally want this multi-root feature!

LennyPenny:
@escaton I can't seem to open the coverage ui
When I use the jest toggle coverage action I can see no noticeable change in the ui

escaton:
@LennyPenny are you talking about "master" (from vscode store)? If not, please try it.

orta:
👋 alright, I've given this a test run on a few projects and it all seems to be working well

@connectdotz wanna give the code another look over then we can get this shipped?

LennyPenny:
@escaton yes I'm using the store version

connectdotz:
things are pretty crazy here after the holiday... sorry for the delay, but I will definitely find some time this weekend, if not sooner...

orokanasaru:
Given that this has already been sitting here for months, and that the current store version of the extension does not work at all in the multi-root scenario, is it really worth continuing to hold off on this merge for some status bar and documentation tweaks?

connectdotz:
I agree this PR is quite useful, and most issues mentioned are indeed not blockers... except one, the status bar reporting, which is unpredictable and really is reporting incomplete/incorrect information... IMHO, If there is anything worse than "not working", it's "not knowing if it is working"...

@escaton will you be able to address the status bar issue? Let me know if you need any help. We are all eagerly awaiting this PR...

escaton:
Let’s first decide whats wrong with status bar. There are tests for new behaviour, did you check it?
My idea was to show most important status at one time, are you ok with this?
If yes but you think it works unpredictable then i probably miss something, please help me to figure out what exactly. I’ll try to fix it quickly.
If no, than unfortunately i dont have enough time for now to reimplement status bar entirely.

src/StatusBar.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+// The bottom status bar
+export class StatusBar {
+  private statusBarSpinner: NodeJS.Timer
+  private statusBarItem = window.createStatusBarItem(StatusBarAlignment.Left)
+  private statusKey = 'Jest:'
+  private frame = elegantSpinner()
+  private priorities: Status[] = ['running', 'failed', 'success', 'stopped', 'initial']
+  private queue: QueueItem[] = []
```
connectdotz: what do users expect from status bar? 
1. the overall project status 
2. the current workspace status. 

For a single workspace project, these status are the same; for multi-workspace, they are different. The current implementation reported what it considered the most _important_ status (by type), which doesn't guarantee to answer either question above. The good news is I think it can be easily addressed without much structure change outside of this file.

1. instead of having a single queue, we can have a dictionary: `{[source]: QueueItem}`,  which holds the latest status of each workspace. 
2. this status dictionary can be used to report project-wide status, such as `running:1, success:3, fail:2` etc, to save space, we should try to replace the text with icon. 
3. this status dictionary can also be used to report "current" workspace status, in the bind() method, you can add a new `setActive()` for example, to display the current workspace's status, in addition to the project-wide status above. `statusBar.setActive()` can be invoked in `JestExt.DidChangeActiveTextEditor()`, for example. So the final status could look like: `Jest: current: 1 test suite failed (all: success:2, fail:1)`...
4. we could even consider creating 2 `statusBarItem` for active workspace and project-wide status, each is associated with its own `statusBarCommand` showing corresponding output channel. The project-level channel would simply report the detail of the whole status dictionary.

This should simplify the logic (no need for priority, queuing etc) a bit while providing more relevant and complete information, compatible with the current implementation. what do you think?

connectdotz:
hey @escaton  have you given up on this PR?

src/StatusBar.ts
kresli: @connectdotz  Is this really a block for merging this PR? This sounds more like improvement because @escaton solution works. I personally never look at status bar for jest extension status, however I'm using `debug` single test feature a lot which currently doesn't work at all in multi root. Can we merge this and create a new ticket with your suggestions?

escaton: I believe that displaying
```
Jest: current: 1 test suite failed (all: success:2, fail:1)
```
is far from good UX. There is not much space in status bar to give it's 1/4 (on my mbp) to one extension.
And also
>Why are we so sure that a failure status from other workspace is more important than the workspace users are currently working on?

Why should we separate code to workspaces in this case? It's all one project with one common status. Your changes in one workspace could potentially broke tests in another workspace. Why should we still message you "Success" in the first place?

Anyway i'm +1 to discuss it in separate issue.

escaton:
I'd prefer to improve status bar in separate issue. But if you think it's a blocker, feel free to close this PR and open another one with your fixes for status bar.

src/StatusBar.ts
connectdotz: sorry for the delay, just returned from a long overseas trip. 

@escaton I see where you come from... you are saying that regardless of single or multiple workspaces, if they are under the same project, we should just display a single status; but given each workspace does report individual status, you had to "hand pick" one thus the priority: `'running' >  'failed' > 'success' > 'stopped' > 'initial'`. 

> Your changes in one workspace could potentially broke tests in another workspace

True, but the other way could also be true: imagine you are working on your own workspace X, but the project shows failure status from other workspace Y that you know has nothing to do with your change. Because the status bar will always show failure from Y, You will not know if your change has caused new test failures (in X) or corrected existing test failures (in X)...  In such a scenario, you can probably understand why the status bar is either useless or confusing, especially when other users are not as familiar with the priority logic as you...

As a tool builder, we are humbled by vast use cases our users presented. Therefore, we always strive to provide a solution with as few assumptions as possible so they can work with more use cases.

> ...is far from good UX. There is not much space in status bar to give its 1/4 (on my mbp) to one extension.

probably, then let's see how we can improve the UX, without compromising the principle we discussed above 😄 ...

LennyPenny:
or even better, make a pr to this branch :D

connectdotz:
I will merge this PR to a beta branch, so others who are interested can contribute and enhance the status logic before we release it.  

@escaton thanks for your hard work, sorry it took a long time. I will update the status later when the merge is done.

src/StatusBar.ts
escaton: > but the project shows failure status from other workspace Y that you know has nothing to do with your change. Because the status bar will always show failure from Y, You will not know if your change has caused new test failures (in X) or corrected existing test failures (in X)

This sounds like you may want to disable extension for that workspace. Since you don't know what to do with that failures.

escaton:
@connectdotz The only thing i might want to improve is reduce settings from `(enabled,disabled)WorkspaceFolders` to just one `enabledWorkspaceFolders` or `disabledWorkspaceFolders`. Because the result of setting both might be confusing. What do you think?

connectdotz:
> reduce settings from (enabled,disabled)WorkspaceFolders to just one enabledWorkspaceFolders or disabledWorkspaceFolders. 

good idea

src/StatusBar.ts
connectdotz: > This sounds like you may want to disable extension for that workspace. Since you don't know what to do with those failures.

you could, but wouldn't that be hacky that you will have to update the workspaces settings based on the change you are working on? What if you also need to work on workspace Z, which does impact Y,  you will have to switch the settings again?

escaton:
you mean handle `disabledWorkspaceFolders` (yeah i think this one should be the only one) here https://github.com/jest-community/vscode-jest/pull/379/files#diff-16a840493184a486054ad1419976b814R121 ?

src/StatusBar.ts
escaton: no, i don't mean you should change project settings once you switch to other code. I mean that if you can't handle errors in Y than maybe you want to mute them all by disabling jest in Y forever.

src/extensionManager.ts
```diff
@@ -0,0 +1,189 @@
...
+  onDidChangeConfiguration(e: vscode.ConfigurationChangeEvent) {
+    if (e.affectsConfiguration('jest')) {
+      const updatedSettings = getExtensionWindowSettings()
+      this.debugCodeLensProvider.showWhenTestStateIn = updatedSettings.debugCodeLens.enabled
+        ? updatedSettings.debugCodeLens.showWhenTestStateIn
+        : []
+    }
+    vscode.workspace.workspaceFolders.forEach(workspaceFolder => {
+      const jestExt = this.getByName(workspaceFolder.name)
+      if (jestExt && e.affectsConfiguration('jest', workspaceFolder.uri)) {
```
connectdotz: should we register/unregister workspace when settings such as `jest.disabledWorkspaceFolders` change? Looks like the `jestExt` was not changed when I add/remove the workspace via the setting`jest.disabledWorkspaceFolders`

src/StatusBar.ts
connectdotz: > if you can't handle errors in Y than maybe you want to mute them all by disabling jest in Y forever.

But you can't disable Y **forever** because sometimes you need to work on Z, which could impact Y; therefore Y should not be disabled in that situation, right? You can probably see the conflict and overhead this approach could impose when working on a larger project or multiple folders... 

Don't give up on your original argument for the importance of the whole project status; it is not wrong, just incomplete for some use cases -- We probably do not want to exclude Y because it is indeed part of the project that we ultimately care. 

The examples above merely points out sometimes we do need to know the status of the workspace one is working on, **in addition to** the project level status. If we allow ourselves to take a step back and question our assumption: why single status? What not display both? Wouldn't that support more use cases with less assumption? "But the status bar has limited space!" sure, so the problem changed from **"how do we choose one status for all use cases"**  to **"how to display both statuses in limited space"** , a much simpler problem that I am confident we will find a satisfying solution quickly.

I understand you might not have time to address this issue, that is quite all right. Once we merge this PR into a branch, others can start to contribute. You have established a great starting point, and we will get this released soon.

connectdotz:
Merged this PR to `beta` branch and created a new issue #442 to track progress. For people who care deeply about this feature, now you can help!

PR #381, opened 2018-09-19
Add sourcemaps for jest

escaton:
This PR fixes sourcemaps for `vscode-jest` itself.
It mostly affects coverage report, but also debugging tests.

Before:
изображение

After:
изображение


connectdotz:
@escaton can you provide more info about the problem this PR is trying to address? thx.

escaton:
The coverage drop is expected. If you check reports from this branch and master side-by-side, you'll see its much more relevant now.
For example: [old report](https://coveralls.io/builds/23636446/source?filename=src/statusBar.ts) vs [new one](https://coveralls.io/builds/23760462/source?filename=src%2FstatusBar.ts)

escaton:
Oh, i see you've already done this in #447 :)

PR #386, opened 2018-10-02
Update Create React App to version 2.x

tristola:


stephtr:
@orta Do we still need to have the CRA sample in this repository? Especially since CRA should be supported out of the box.

orta:
This was mainly so I had a CRA app to test with, so updating it is harmless IMO

orta:
Thanks @tristola

PR #441, opened 2019-04-04
Restart command

### LINKED ISSUE #59: Add Reset Jest Runner command
I'm setting up  a new project, I'm stopping and starting all the time

vdh:
- Added a "Jest: Restart Runner" / `restartProcess` command that calls `stopProcess` then `startProcess`
- Updated `stopAll` in the process manager to return a promise (via `Promise.all`), similar to `stopJestProcess`
- Updated `stopProcess` to return the promise from `stopAll`
- To keep things DRY, replaced some restart logic at the end of `triggerUpdateSettings` with a call to `restartProcess`, and used the promise from `stopProcess` instead of the 500ms timeout

Fixes #59

### DIFF
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -215,7 +215,7 @@ export class JestExt {
   }
 
   public restartProcess() {
-    this.stopProcess().then(() => {
+    return this.stopProcess().then(() => {
       this.startProcess()
     })
   }
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -417,7 +417,7 @@ export class JestExt {
 
   public runTest = async (workspaceFolder: vscode.WorkspaceFolder, fileName: string, identifier: string) => {
     const restart = this.jestProcessManager.numberOfProcesses > 0
-    this.jestProcessManager.stopAll()
+    await this.jestProcessManager.stopAll()
 
     this.debugConfigurationProvider.prepareTestRun(fileName, identifier)
 
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -213,6 +213,14 @@ export class JestExt {
     this.status.stopped()
   }
 
+  public restartProcess() {
+    this.stopProcess()
+
+    setTimeout(() => {
+      this.startProcess()
+    }, 500)
+  }
+
   private detectedSnapshotErrors() {
     if (!this.pluginSettings.enableSnapshotUpdateMessages) {
       return
@@ -267,11 +275,7 @@ export class JestExt {
 
     this.coverageOverlay.enabled = updatedSettings.showCoverageOnLoad
 
-    this.stopProcess()
-
-    setTimeout(() => {
-      this.startProcess()
-    }, 500)
+    this.restartProcess()
   }
 
   updateDecorators(testResults: SortedTestResults, editor: vscode.TextEditor) {
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -209,16 +209,15 @@ export class JestExt {
 
   public stopProcess() {
     this.channel.appendLine('Closing Jest')
-    this.jestProcessManager.stopAll()
-    this.status.stopped()
+    return this.jestProcessManager.stopAll().then(() => {
+      this.status.stopped()
+    })
   }
 
   public restartProcess() {
-    this.stopProcess()
-
-    setTimeout(() => {
+    this.stopProcess().then(() => {
       this.startProcess()
-    }, 500)
+    })
   }
 
   private detectedSnapshotErrors() {
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -417,7 +417,7 @@ export class JestExt {
 
   public runTest = async (workspaceFolder: vscode.WorkspaceFolder, fileName: string, identifier: string) => {
     const restart = this.jestProcessManager.numberOfProcesses > 0
-    await this.jestProcessManager.stopAll()
+    this.jestProcessManager.stopAll()
 
     this.debugConfigurationProvider.prepareTestRun(fileName, identifier)
 
--- a/src/JestProcessManagement/JestProcess.ts
+++ b/src/JestProcessManagement/JestProcess.ts
@@ -72,13 +72,13 @@ export class JestProcess {
     this.onExitCallback = callback
   }
 
-  public onJestEditorSupportEvent(event, callback) {
+  public onJestEditorSupportEvent(event: string, callback: (...args: any) => void) {
     this.jestSupportEvents.set(event, callback)
     this.runner.on(event, callback)
     return this
   }
 
-  public stop() {
+  public stop(): Promise<undefined> {
     this.stopRequested = true
     this.keepAliveCounter = 1
     this.jestSupportEvents.clear()
@@ -88,7 +88,7 @@ export class JestProcess {
     })
   }
 
-  public runJestWithUpdateForSnapshots(callback) {
+  public runJestWithUpdateForSnapshots(callback: Function) {
     this.runner.runJestWithUpdateForSnapshots(callback)
   }
 }
--- a/src/JestProcessManagement/JestProcess.ts
+++ b/src/JestProcessManagement/JestProcess.ts
@@ -4,6 +4,7 @@ import { WatchMode } from '../Jest'
 
 export class JestProcess {
   static readonly keepAliveLimit = 5
+  static readonly stopHangTimeout = 500
   private runner: Runner
   private projectWorkspace: ProjectWorkspace
   private onExitCallback: Function
@@ -79,12 +80,14 @@ export class JestProcess {
   }
 
   public stop(): Promise<void> {
-    this.stopRequested = true
-    this.keepAliveCounter = 1
-    this.jestSupportEvents.clear()
     return new Promise(resolve => {
+      this.stopRequested = true
+      this.keepAliveCounter = 1
       this.resolve = resolve
+      this.jestSupportEvents.clear()
       this.runner.closeProcess()
+
+      setTimeout(resolve, JestProcess.stopHangTimeout)
     })
   }
 
--- a/src/JestProcessManagement/JestProcess.ts
+++ b/src/JestProcessManagement/JestProcess.ts
@@ -72,13 +72,13 @@ export class JestProcess {
     this.onExitCallback = callback
   }
 
-  public onJestEditorSupportEvent(event: string, callback: (...args: any) => void) {
+  public onJestEditorSupportEvent(event: string, callback: (...args: any[]) => void) {
     this.jestSupportEvents.set(event, callback)
     this.runner.on(event, callback)
     return this
   }
 
-  public stop(): Promise<undefined> {
+  public stop(): Promise<void> {
     this.stopRequested = true
     this.keepAliveCounter = 1
     this.jestSupportEvents.clear()
@@ -88,7 +88,7 @@ export class JestProcess {
     })
   }
 
-  public runJestWithUpdateForSnapshots(callback: Function) {
+  public runJestWithUpdateForSnapshots(callback: () => void) {
     this.runner.runJestWithUpdateForSnapshots(callback)
   }
 }
--- a/src/JestProcessManagement/JestProcess.ts
+++ b/src/JestProcessManagement/JestProcess.ts
@@ -82,9 +82,9 @@ export class JestProcess {
     this.stopRequested = true
     this.keepAliveCounter = 1
     this.jestSupportEvents.clear()
-    this.runner.closeProcess()
     return new Promise(resolve => {
       this.resolve = resolve
+      this.runner.closeProcess()
     })
   }
 
--- a/src/JestProcessManagement/JestProcess.ts
+++ b/src/JestProcessManagement/JestProcess.ts
@@ -39,6 +39,7 @@ export class JestProcess {
           this.onExitCallback(this)
           if (this.stopRequested) {
             this.resolve()
+            this.resolve = null
           }
         }
       }
@@ -87,7 +88,13 @@ export class JestProcess {
       this.jestSupportEvents.clear()
       this.runner.closeProcess()
 
-      setTimeout(resolve, JestProcess.stopHangTimeout)
+      // As a safety fallback to prevent the stop from hanging, resolve after a timeout
+      // TODO: If `closeProcess` can be guarenteed to always resolve, remove this
+      setTimeout(() => {
+        if (this.resolve === resolve) {
+          resolve()
+        }
+      }, JestProcess.stopHangTimeout)
     })
   }
 
--- a/src/JestProcessManagement/JestProcessManager.ts
+++ b/src/JestProcessManagement/JestProcessManager.ts
@@ -18,14 +18,22 @@ export class JestProcessManager {
     this.runAllTestsFirstInWatchMode = runAllTestsFirstInWatchMode
   }
 
-  private removeJestProcessReference(jestProcess) {
+  private removeJestProcessReference(jestProcess: JestProcess) {
     const index = this.jestProcesses.indexOf(jestProcess)
     if (index !== -1) {
       this.jestProcesses.splice(index, 1)
     }
   }
 
-  private runJest({ watchMode, keepAlive, exitCallback }) {
+  private runJest({
+    watchMode,
+    keepAlive,
+    exitCallback,
+  }: {
+    watchMode: WatchMode
+    keepAlive: boolean
+    exitCallback: Function
+  }) {
     const jestProcess = new JestProcess({
       projectWorkspace: this.projectWorkspace,
       watchMode,
@@ -38,7 +46,15 @@ export class JestProcessManager {
     return jestProcess
   }
 
-  private run({ watchMode, keepAlive, exitCallback }) {
+  private run({
+    watchMode,
+    keepAlive,
+    exitCallback,
+  }: {
+    watchMode: WatchMode
+    keepAlive: boolean
+    exitCallback: Function
+  }) {
     return this.runJest({
       watchMode,
       keepAlive,
@@ -51,7 +67,7 @@ export class JestProcessManager {
     })
   }
 
-  private runAllTestsFirst(onExit) {
+  private runAllTestsFirst(onExit: Function) {
     return this.runJest({
       watchMode: WatchMode.None,
       keepAlive: false,
@@ -95,7 +111,7 @@ export class JestProcessManager {
     return Promise.all(processesToRemove.map(jestProcess => jestProcess.stop()))
   }
 
-  public stopJestProcess(jestProcess) {
+  public stopJestProcess(jestProcess: JestProcess) {
     this.removeJestProcessReference(jestProcess)
     return jestProcess.stop()
   }
--- a/src/JestProcessManagement/JestProcessManager.ts
+++ b/src/JestProcessManagement/JestProcessManager.ts
@@ -92,9 +92,7 @@ export class JestProcessManager {
   public stopAll() {
     const processesToRemove = [...this.jestProcesses]
     this.jestProcesses = []
-    processesToRemove.forEach(jestProcess => {
-      jestProcess.stop()
-    })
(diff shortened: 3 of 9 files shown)


vdh:
Phew, okay, those typos during that rebase were embarrassing, but the changes have now been rebased to `beta` instead of `master`. The old version of the branch is available at https://github.com/vdh/vscode-jest/tree/restart-command-from-master for reference.

vdh:
@escaton as you mentioned on #443, I've rebased it to the `beta` branch.

Since this is my first time contributing to `vscode-jest`, I'd be grateful if someone could review whether or not the promise-based changes I've made are safe.

The `stopAll` and `restartProcess` changes are simple enough, but it would be great to get confirmation if:
- Returning a promise from `stop` in `JestProcess.ts` is safe? The callback of `this.resolve()` is inside the event handler for `debuggerProcessExit`. Is it safe to assume that all `JestProcess` instances will be sent a `debuggerProcessExit` event after calling `this.runner.closeProcess()`?
- The `await` for `stopAll` in `runTest` is a good fix to avoid a race condition? I assumed it would be good to wait for the stop to finish before continuing.

vdh:
Oh, to clarify what I meant by "Returning a promise", is that currently the existing `stop`'s promise is only used in `stopJestProcess`, so is it safe to start depending on it for `stopAll`. (i.e. if it doesn't always get resolved, the `stopAll()` might hang)

src/JestExt.ts
```diff
@@ -267,11 +274,7 @@ export class JestExt {
 
     this.coverageOverlay.enabled = updatedSettings.showCoverageOnLoad
 
-    this.stopProcess()
-
-    setTimeout(() => {
```
connectdotz: I think the runner should report 'debuggerProcessExit event' in the normal situation (multiple times perhaps), but judging the fact that there is a timeout watch makes me wonder if there was incidence that it didn't... 

do you think it make sense to add the timer watch to `JestProcess` (not here) in case the stop promise didn't get resolved accordingly? If we are concering about the overhead, maybe add a extension option to control this behavior so users can turn it on when such rare situations occur...?

src/JestProcessManagement/JestProcessManager.ts
```diff
@@ -92,9 +92,7 @@ export class JestProcessManager {
   public stopAll() {
     const processesToRemove = [...this.jestProcesses]
     this.jestProcesses = []
-    processesToRemove.forEach(jestProcess => {
-      jestProcess.stop()
-    })
+    return Promise.all(processesToRemove.map(jestProcess => jestProcess.stop()))
```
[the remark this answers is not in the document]
vdh: The existing code is already resetting `this.jestProcesses` before attempting to stop them. I assume this is so consecutive calls don't overlap on attempts to stop the same process instance? From what I can see, `stopAll` has basically the same behaviour as `stopJestProcess`.

src/JestProcessManagement/JestProcess.ts
```diff
@@ -82,9 +82,9 @@ export class JestProcess {
     this.stopRequested = true
     this.keepAliveCounter = 1
     this.jestSupportEvents.clear()
-    this.runner.closeProcess()
     return new Promise(resolve => {
       this.resolve = resolve
```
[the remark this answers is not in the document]
connectdotz: additional note: the stop-process efficiency will become critical once we switch to the promise-base start/stop, thus we need to ensure there is no loose logic or inefficiency...

src/JestExt.ts
vdh: The original message for 13921d74b05069c85bb9abdfb661b827af0f4046 doesn't read like it was specific to needing a timeout, so any (possible) dependencies on the timeout would probably be unintentional?

So are you thinking there should be a `setTimeout(resolve)` or a `Promise.race` in the `stop` method of `JestProcess`, as a fallback in case the stop hangs?

src/JestProcessManagement/JestProcess.ts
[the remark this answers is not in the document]
vdh: If I'm reading the logic of `debuggerProcessExit` right, the `--this.keepAliveCounter` should decrement the previously-set `1` to `0` and skip the `if` that would try to restart the process (unless there's an edge case where `keepAliveCounter` is modified again somewhere between the `stop` call and `debuggerProcessExit` firing…?).

src/JestExt.ts
connectdotz: hmmm... it seems to use a timer for rare edge case is not ideal... let's make sure the stopProcess will terminate the promise if it is still running (and I assume we will call stopProcess for every startProcess?). This way users can terminate the process themselves should it happen occasionally. what do you think?

src/JestProcessManagement/JestProcess.ts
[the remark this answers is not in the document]
connectdotz: I think you are right, but something seems convoluted...  the whole `keepAlive` and `keepAliveCounter` are unnecessary red herrings, which can probably be simplified:

1. don't think `keepAlive` needs to be an ivar, it was never used outside of the constructor.
1. inside the `startRunner()`, do we need both `exited` and `keepAliveCounter`?  when `keepAliveCounter` reaches 0, doesn't it mean exited?
1. does `keepAliveCounter` need to be an ivar? if we want to force-stop the process, isn't there a "stopRequested` ivar? It seems we won't need `keepAliveCounter` outside of the `startRunner()`, so the code would be a lot cleaner if we just pass it in like `startRunner(kepAliveCounter)`, a pretty standard self-captured count-down pattern. It should check `this.stopRequested` when deciding if to recursively calling itself with the decremented counter.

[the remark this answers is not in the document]
vdh: I think `exited` exists to prevent the `onExitCallback` from being called more than once, on the edge case where `debuggerProcessExit` is called more than once? I agree though that the flow is not ideal and could be refactored with fewer variables. The `keepAliveCounter = 1` hack could be replaced with `this.stopRequested` having a higher priority in the listener than the `keepAlive` logic.

The `keepAlive` iVar seems to be used (read-only?) in `JestProcessManager`. It looks like @marcinczenko would know more about why it exists since he's the one who originally made it.

src/JestExt.ts
vdh: How would the force kill be accessed through the UI? Also, resolving the promise would only fix `stopAll` hanging, it wouldn't actually do any "killing" without additional logic.

src/JestProcessManagement/JestProcess.ts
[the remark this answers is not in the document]
marcinczenko: When I was rewriting the process logic, I was also facing backward compatibility logic. If I remember correctly `keepAlive` is one of them. `startJestProcess` is always called with `keepAlive` being set to `true`, but for some reason at the time of creating it, we decided to have it as an option. Perhaps it can be factored out, which may remove some backward compatibility noise. As you can see, it is used only in the constructor of `JestProcess` to force `keepAliveLimit` to be `1` in case `keepAlive` argument is set to `false`.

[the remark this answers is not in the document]
marcinczenko: The promise in `JestProcess.stop` has been added at the very end of writing the process logic. Its intention was to support restarting Jest after updating the snapshots:

```javascript
  private detectedSnapshotErrors() {
    if (!this.pluginSettings.enableSnapshotUpdateMessages) {
      return
    }
    vscode.window
      .showInformationMessage('Would you like to update your Snapshots?', { title: 'Replace them' })
      .then(response => {
        // No response == cancel
        if (response) {
          this.jestProcess.runJestWithUpdateForSnapshots(() => {
            if (this.pluginSettings.restartJestOnSnapshotUpdate) {
              this.jestProcessManager.stopJestProcess(this.jestProcess).then(() => {
                this.startProcess()
              })
              vscode.window.showInformationMessage('Updated Snapshots and restarted Jest.')
            } else {
              vscode.window.showInformationMessage('Updated Snapshots. It will show in your next test run.')
            }
          })
        }
      })
  }
```

The promise logic was not added to `stopAll` because `stopAll` was not at that time intended to be used in the *restart* logic and more as "clean after yourself and stop forever" thing. The restart logic should be used in the same way as in the code snippet shown above.

I do not know the true intention of the:

```javascript
this.stopProcess()
setTimeout(() => {
  this.startProcess()
}, 500)
```

sequence, I suspect, that what I just said was not obvious and a higher level API has been used to achieve that. And since the promise logic is not present in `stopAll` hence the timeout.

Making `stopAll` waiting for all promises is in my opinion worth considering as this would make API more consistent. I am not aware of any hanging and I think the timeout was not used for such reason.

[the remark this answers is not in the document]
connectdotz: thanks @marcinczenko for good background info update. 

from the code snippet:
```
             this.jestProcessManager.stopJestProcess(this.jestProcess).then(() => {
                this.startProcess()
              })
```
I wonder if you experienced issues when `startProcess()` is called without waiting for `stopJestProcess()`? I guess I am curious the underlying issue the promised-based stop() is trying to resolve, giving the jestProcess is not reused in the subsequent start()...

src/JestExt.ts
connectdotz: > How would the force kill be accessed through the UI

same as how we introduced new start/stop-all commands

>  resolving the promise would only fix stopAll hanging, it wouldn't actually do any "killing" without additional logic.

"fix hanging" is what "force kill" aim to do. Don't think we can do better in terms of "killing the process" after calling the `stop()`. All we try to avoid is a hanging promise preventing the subsequent start of new jestProcess.

However, I am having a second thought about the timeout, given `stop()` is only called during specific user interactions, such as users issue "Jest: Stop Runners" command, changing settings or update snapshots, i.e. it is not frequent. In that context, I think to have a timeout watching the zombie promise in `stop()` might be acceptable...

But all these might be moot points, depends on the underlying reason of the promise,  giving 1. we don't reuse the jestProcess in subsequent start() 2. there is nothing we attempt to do if it didn't exit as expected...  I can't help but wonder what the consequence for a  much simpler "kill-and-forget" pattern is? (i.e. the question I asked @marcinczenko above)

src/JestProcessManagement/JestProcess.ts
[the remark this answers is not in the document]
marcinczenko: I do not recall having any specific issues. I remember that there was some conversations about making all stuff more and more synchronous, so this was maybe the first step towards this. I remember I first implemented restarting without promise logic - it was added later and must have been influenced by the reviews. As well pointed at that times, leaning towards more synchronous behaviour lowers complexity and that was I think the main trigger. It may also have been that in this specific case, keeping the UI updates more synchronous is nicer user experience (but I am no longer sure if this was a factor).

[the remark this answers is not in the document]
connectdotz: yes, I vaguely remember the conversation about making the jest process more sequential with Promise. Ok, so that confirmed we didn't create a delay for any specific issue but just following the best practice.

src/JestProcessManagement/JestProcess.ts
```diff
@@ -72,13 +72,13 @@ export class JestProcess {
     this.onExitCallback = callback
   }
 
-  public onJestEditorSupportEvent(event, callback) {
+  public onJestEditorSupportEvent(event: string, callback: (...args: any) => void) {
```
connectdotz: I think it's better to use array type for rest parameter: `(...args: any[])`

src/JestProcessManagement/JestProcess.ts
```diff
@@ -88,7 +88,7 @@ export class JestProcess {
     })
   }
 
-  public runJestWithUpdateForSnapshots(callback) {
+  public runJestWithUpdateForSnapshots(callback: Function) {
```
connectdotz: can we be more explicit about the function signature?

src/JestProcessManagement/JestProcess.ts
```diff
@@ -72,13 +72,13 @@ export class JestProcess {
...
 
-  public onJestEditorSupportEvent(event, callback) {
+  public onJestEditorSupportEvent(event: string, callback: (...args: any) => void) {
     this.jestSupportEvents.set(event, callback)
     this.runner.on(event, callback)
     return this
   }
 
-  public stop() {
+  public stop(): Promise<undefined> {
```
connectdotz: do you mean `Promise<void>`?

connectdotz: do you mean `Promise<void>`?

src/JestProcessManagement/JestProcess.ts
connectdotz: can we be more explicit about the function signature?

src/JestProcessManagement/JestProcess.ts
connectdotz: it's better to use an array signature for the rest argument: `(...args: any[])`

src/JestExt.ts
connectdotz: let's try to close this thread, my 2 cents:
1. JestProcess can undoubtedly be cleaned up and streamlined, and hopefully, we can one day move to full Promise model (for the "start") instead of recursive invocation and complex state management... But it could be done in a separate PR.
1. regarding "stop", a promise is fine as we can still implement kill-and-forget in the timeout watch anyway. However, I struggled between a more forgiving approach (with an additional timeout to prevent zombie process blocking the subsequent start) and a more strict approach  (if the process didn't report close for any reason, no subsequent jest could be started). Considering the principle of _"do not block"_, I am leaning toward the more forgiving approach, which is also more compatible with the current behavior...

@vdh does that give you enough to proceed?

src/JestProcessManagement/JestProcess.ts
vdh: Ah yes sorry, this was my first foray into Typescript, and I was unsure how to type the existing empty-resolve Promise.

src/JestProcessManagement/JestProcess.ts
vdh: `JestExt` uses it with an empty-parameter callback, and `Runner` is even looser with a `runJestWithUpdateForSnapshots(completion: any): void` signature. Internally, it calls it as `completion();`. So if you have advice on a better signature I'm all ears (I'm relatively new to Typescript).

src/JestExt.ts
vdh: @connectdotz So if I'm interpreting things correctly, we should avoid refactoring `JestProcess` for now in the scope of this PR, and just add some additional logic to prevent blocking? Would this change be suitable?

Remove the `await` inside `runTest` (revert my addition to that line) to avoid blocking there, and add a `Promise.race` to `stopProcess`?

```ts
export class JestExt {
  // …
  public async stopProcess() {
    this.channel.appendLine('Closing Jest')
    await Promise.race([
      this.jestProcessManager.stopAll(),
      new Promise((resolve) => { setTimeout(resolve, 500) }),
    ])
    this.status.stopped()
  }
  // …
}
```

I feel like the magic number on the timeout probably needs more refinement or labelling though.

The alternative would be to add a race inside `stopAll`, but then that would mean `stopJestProcess` would need changes as well, and that could get messy…?

connectdotz: sorry for the late reply, can we encapsulate this stop logic inside the JestProcess itself?

connectdotz: I should add this doesn't mean we need to refactor the rest of the JestProcess, I am thinking this can be easily encapsulated inside the JestProcess's `stop()` method without touching the rest.

src/JestProcessManagement/JestProcess.ts
connectdotz: in `jest-editor-support`:
```
runJestWithUpdateForSnapshots(completion: any, args: string[]) 
```
indeed the completion signature is too loose, it should have been `() => void`, and the `args` actually is optional, I will fix it on the jest-editor-support side. As for this PR, I think we should do the right thing by declaring  
```
public runJestWithUpdateForSnapshots(callback: () => void) 
```

Note: a bit background for changing 'Function': tslint recommendation [ban-types](https://github.com/palantir/tslint/blob/3efdfe57b7c0eb1b21e21cec128c993cf20da438/src/configs/recommended.ts#L28)

vdh: Hm, I installed the TSLint extension I was missing, but the project doesn't seem configured to hint for that rule at the moment. Is there something I'm missing, or is this just advice based on your previous experience with that rule?

src/JestExt.ts
vdh: I've pushed an update with a `setTimeout` inside the `stop()` promise

src/JestProcessManagement/JestProcess.ts
connectdotz: we probably should have added tslint sooner, but at least it will be there after #447, which I have already fixed a lot of `Function` types... 😉

src/JestProcessManagement/JestProcess.ts
```diff
@@ -79,12 +80,14 @@ export class JestProcess {
...
-    this.keepAliveCounter = 1
-    this.jestSupportEvents.clear()
     return new Promise(resolve => {
+      this.stopRequested = true
+      this.keepAliveCounter = 1
       this.resolve = resolve
+      this.jestSupportEvents.clear()
       this.runner.closeProcess()
+
+      setTimeout(resolve, JestProcess.stopHangTimeout)
```
connectdotz: this looks clean, however, for the sake of future maintenance,  I am thinking we should
1. add some comment to brief why we are adding this timeout watch here
1. If we are serious about removing this in the future, we should have a way to tell if the timeout is "redundant", i.e. the promise has already been resolved.  Right now it doesn't seem to be able to tell us that. We can certainly refactor it to reset resolve, emit logging etc, but it will no doubt adding more pain for the already confusing state management here... hmmm... Let's at least add a TODO comment to outline what should be done, or create an issue to track all the issues we discussed here but deferred?

BTW, I assume you have build vscode-jest with this PR and tested in your dev environment, right? I haven't got a chance to do so, if you have done that then we can merge once the comment is added.

Again, thanks for your contribution 👍 , sorry for the delay but we are almost there...

vdh: 1. Do you have advice for how I should word the comment?
2. From the what I've found, [calling resolve additional times should be a no-op](https://stackoverflow.com/questions/20328073/is-it-safe-to-resolve-a-promise-multiple-times) ("If alreadyResolved.[[value]] is true, return undefined."). But if we wanted to avoid calling it anyway, we could perhaps set `this.resolve = null` after the normal `resolve()` and then wrap the timeout's resolve in a `if (this.resolve === resolve)` wrapper?

I have done some brief tests with running it in my dev environment, but I haven't done a serious amount of dogfooding with it yet.

connectdotz: > Do you have advice for how I should word the comment?

put yourself in the future developers' shoes, when they see the promise is resolved in the callback functions, and then find a timeout watch here, wouldn't they be confused? Remeber the question we were asking when we saw the original timeout in JestExt?... I suggest adding  the "why", and a TODO to outline your ultimate vision... oh and feel free to link back to the relevant discussions in this PR, if needed.

> But if we wanted to avoid calling it anyway, we could perhaps set this.resolve = null after the normal resolve() and then wrap the timeout's resolve in a if (this.resolve === resolve) wrapper?

something like that, but giving there are already too many questionable ivar/state that we want to clean up, I really hate to see us piling up more... that's why I am open to tackle these state-management related issues in a separate refactoring PR, so long as we put some comments or file an issue to follow up...

vdh: I've added some comments about the timeout, and the `this.resolve === resolve` sanity check. Hopefully, a future refactor can remove the need for the `this.resolve` ivar altogether, and make it more promise-based.

connectdotz:
@vdh the test coverage has degraded, can you address it? thx.

connectdotz:
hey @vdh do you need some help?

vdh:
@connectdotz Sorry for the delay, I've been pressed for time the past few days. Help would be much appreciated if it's not too much trouble for you.

escaton:
I could take care of coverage.

PR #443, opened 2019-04-13
Multiroot improvements

escaton:
- [x] ~add start-all, stop-all, restart{-all} commands~
- [x] handle `disabledWorkspaceFolders` change on the fly
- [x] misc: use Map to store jest instances

#trivial

### DIFF
--- a/src/extensionManager.ts
+++ b/src/extensionManager.ts
@@ -91,7 +91,8 @@ export class ExtensionManager {
     const { commonPluginSettings: { disabledWorkspaceFolders } } = this
     if (this.extByWorkspace.has(workspaceFolderName)) {
       return false
-    } else if (disabledWorkspaceFolders.includes(workspaceFolderName)) {
+    }
+    if (disabledWorkspaceFolders.includes(workspaceFolderName)) {
       return false
     }
     return true
--- a/src/extensionManager.ts
+++ b/src/extensionManager.ts
@@ -16,16 +16,16 @@ export class ExtensionManager {
 
   constructor(context: vscode.ExtensionContext) {
     this.context = context
-
-    this.commonPluginSettings = getExtensionWindowSettings()
-
-    this.debugCodeLensProvider = new DebugCodeLensProvider(
-      uri => this.getByDocUri(uri),
-      this.commonPluginSettings.debugCodeLens.enabled ? this.commonPluginSettings.debugCodeLens.showWhenTestStateIn : []
-    )
     this.debugConfigurationProvider = new DebugConfigurationProvider()
-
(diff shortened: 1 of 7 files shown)


escaton:
@connectdotz please take a look.
Unfortunately i noticed too late that there was #441
@vdh you've done great research and i would love to drop my commit in favor of your solution, but it would require some refactor from you. I'd suggest you to rebase to `beta` branch and check out the changes.

src/extensionManager.ts
```diff
@@ -53,42 +54,51 @@ export class ExtensionManager {
...
-    return (
-      // start if enabledWorkspaceFolders contains current workspace
-      enabledWorkspaceFolders.includes(workspaceFolderName) ||
-      // dont start if disabledWorkspaceFolders doesn't contain current workspace
-      !disabledWorkspaceFolders.includes(workspaceFolderName)
-    )
+    const { commonPluginSettings: { disabledWorkspaceFolders } } = this
+    if (this.extByWorkspace.has(workspaceFolderName)) {
+      return false
+    } else if (disabledWorkspaceFolders.includes(workspaceFolderName)) {
```
connectdotz: no need for `else`,  you can also consider consolidating these 2 conditions into a single if statement, might be more readable...

connectdotz:
probably update changelog since many users have been asking for these new commands

escaton:
Yes, but i’m considering remove that commit and let @vdh to update his PR. So i guess i’ll add `#trivial`

src/extensionManager.ts
escaton: I did this with the same motivation :)

connectdotz: what is the reason for `else`?

escaton: Ah, missed that. No reason for `else`, i'll split it in two separate conditions.

vdh:
@escaton I've rebased the PR, although since this is my first contribution it'd be great to get someone double check the logic of the Promise-based changes I've made.

escaton:
Dropped my commits related to start/stop/restart commands

Bnaya:
Is there any overview of the features here? 
Thanks ! :)

connectdotz:
> Is there any overview of the features here?

there is a brief introduction in the [README.md](https://github.com/jest-community/vscode-jest/blob/beta/README.md#use-extension-in-multiroot-environment), if it's not intuitive then maybe we should consider adding more...

escaton:
@connectdotz would you like to merge this PR to `beta`?

connectdotz:
yes, was waiting for the restart-command PR, now it is in, is there anything you want to change in this PR before I merge it?

escaton:
I'm done :)

Issue #442, opened 2019-04-12
next release


connectdotz:
This is to track the next release: 

## multi-root workspace support
- [x] merge #379
- [x] fix status bar (#447)
- [x] fix output file generation for concurrent Runner in `jest-editor-support` ([jest-editor-support/#9](https://github.com/jest-community/jest-editor-support/pull/9), #447)
- [x] upgrade `jest-editor-support` dependency to pick up the latest version. (#447)

## sidebar support (postponed)
- <strike>[ ] merge #357 </strike>
- <strike>[ ] support hierarchical tree structure (like the Explorer)</strike>

## additional features
- [x] restart command (#441)
- [x] template-literal-string support (#447)
- [x] fix debug break points problems for some js projects (#459)

## technical debt
- [x] merge #446 (deprecate vscode.previewHtml)
- [x] address security alerts
---
Due to the scope of the changes, we created a `beta` branch so more people can help to address the remaining issues and speed up the release. If you are interested to help, please submit PR against `beta` branch or help us testing. Thank you 🙏 

In addition to testing new features listed above, would really appreciate help for regression testing:
- [ ]  with older vscode (prior to 1.34)?
- [ ]  with older jest (prior to 24.7)?
- [x]  in window env?

connectdotz:
@escaton when I was testing multiroot in beta, noticed this issue: 

The test result parsing has a random race condition, and tests are not being correctly decorated, especially when the system is under a bit of stress. After taking a closer look at the console output, it seems all the jest runners output to the same tmp file! 

```
spawning process with command=/Users/vsun/ConnectDotz/projects/dots/packages/dots-google-photos/node_modules/.bin/jest, args=--json,--useStderr,--outputFile,/var/folders/kv/r07v4crd5qx_kd6959jjdmb80000gn/T/jest_runner.json,--no-color
...
spawning process with command=/Users/vsun/ConnectDotz/projects/dots/packages/dots-message/node_modules/.bin/jest, args=--json,--useStderr,--outputFile,/var/folders/kv/r07v4crd5qx_kd6959jjdmb80000gn/T/jest_runner.json,--no-color
```

[...]

PR #393, opened 2018-10-14
Fix #331: Support useWsl parameter

mojadev:
As discussed, now a version without any mapping in vscode-jest itself. This relies on the changes in the jest-editor-support package, so expect the CI to fail 😃

Issue #247, opened 2018-02-03
Backticks in test case description cause the debugging and monitoring to fail


daelmaak:
### Environment

1. `node -v`: 9.2.0
2. `npm -v`: 5.5.1
3. `npm ls react-scripts` (if you haven’t ejected): empty

4. Operating system: [fill]
Win 10

### Steps to Reproduce

![2018-02-03_21h00_37](https://user-images.githubusercontent.com/34311965/35770971-a7f995f6-0925-11e8-8b2a-dd77d6643182.gif)

1) Have a passing test with single quotes in description 
2) Then changing single quotes to backticks

### Expected Behavior

The status on the left side of the test case is green, debugging works and error messages are provided next to code in case the test is not passing.

### Actual Behavior

The status shows the test as not being run although it runs in the background. If not passing, there are no error messages next to code, only red underline on the culprit row. Debugging doesn't work in this test case.

seanpoulter:
Thanks for reporting the issue @daelmaak.

That's a issue with `jest-editor-support`. There's an open issue for it that just needs some attention and a PR. Have a look at facebook/jest#4910.

daelmaak:
I see you already started with the PR @seanpoulter, are you still working on that?

seanpoulter:
Nope. I stopped with that last comment. If I was going to jump back into it, I'd want to go have a look how the AST handles tagged template literals:
```

jkjustjoshing:
Looks like [this comment](https://github.com/facebook/jest/pull/7215#issuecomment-439109512) indicates that this might just need some version/dependency shuffling to get this working?

connectdotz:
hi, template-literal-string support has been added to [3.0.0-preRelease](https://github.com/jest-community/vscode-jest/releases/tag/3.0.0), feel free to give it a try...

PR #447, opened 2019-04-20
upgrade jest-editor-support and multiroot status

connectdotz:
This PR mainly to 
1. integrate with the latest `'jest-editor-support'`, which contains many fixes needed by multiroot (#442), template-literal(#247), windows useWsl (#393)
1. address multiroot status(#379) 
1. paying some technical debt...

changes by category:

## multiroot
- StatusBar reports both workspace summary and active folder jest status. Added new summary output channel to give full details of each folder status.
- integrate with the latest `jest-editor-support' to pick up concurrency and others fixes from [jest-editor-support/#9](https://github.com/jest-community/jest-editor-support/pull/9)

## misc improvement
- support template-literal-string test names (#247)
  - when failed to match template-literal-string test names, it can now fall back to matching by locations reported by jest.

## technical debt (internal development improvement)
- adopts ts-jest for better sourceMap support and more 
- adopt the latest 'tslint' rules for best practice coding style.
- migrate to typescript 3.x and jest 24.7.x
- some tests clean up to consolidate mocks and others.

---
Note:
- This PR has a dependency on [jest-editor-support/#9](https://github.com/jest-community/jest-editor-support/pull/9), the build will fail until a new release of 'jest-editor-support' is cut.
- there are lots of files touched due to typescript/lint changes, most to conform to typescript 3.x and lint rules.


connectdotz:
Regarding the multiroot-status, I ended up displaying 2 separate statusBarItem - active folder and workspace summary - each command opens up an output channel for more details. 

- The summary status is pretty much @escaton original implement, I just added a new output channel to display all folder's status when clicked. 
- The active folder status is more dynamic, mainly depends on the active editor. The status will open the corresponding output channel when clicked. Unlike single-root, this status will not display the long details strings.

Here is a snapshot since you guys won't be able to build this PR yet (due to jest-editor-support). Please don't hesitate to suggest how we can further improve it...

multiroot-status

(single-root projects status will appear identical to the production version)

connectdotz:
This PR is rather large, maybe we could have multiple people reviewing it?

@escaton, do you mind reviewing the mutiroot related files, such as extensionManager, StatusBar, extension, JestExt? 

@orta or @seanpoulter or @stephtr, I know it is a pain to go through so many files, but many exciting new features are riding on these PRs, when you get a chance, do you mind reviewing both [jest-editor-support/#9](https://github.com/jest-community/jest-editor-support/pull/9) and this one. Thanks.

src/StatusBar.ts
```diff
@@ -1,26 +1,62 @@
...
 
 type Status = 'running' | 'failed' | 'success' | 'stopped' | 'initial'
-type QueueItem = {
+interface QueueItem {
   source: string
   status: Status
   details: string | undefined
 }
 
+interface StatusBarSpinner {
```
escaton: Not the best name for such composite entity. I'd prefer to have two separate spinners, like `StatusItem` below

src/StatusBar.ts
```diff
@@ -1,26 +1,62 @@
...
+    this.summaryStatusItem.command = showSummaryOutput
+    this.activeStatusItem.command = showAciiveOutput
+
+    return [
+      vscode.commands.registerCommand(showSummaryOutput, () => {
+        if (this.workspaceOutput) {
+          this.workspaceOutput.show()
+        }
+      }),
+      vscode.commands.registerCommand(showAciiveOutput, () => {
```
escaton: typo

src/StatusBar.ts
```diff
@@ -41,32 +77,156 @@ export class StatusBar {
...
+      }
+      return
+    }
+
+    // find the active item from the queue
+    const queueItem = this.activeFolder
+      ? this.queue.find(_item => _item.source === this.activeFolder)
+      : this.queue.length === 1
+      ? this.queue[0]
+      : undefined
```
escaton: If this is not an explicit code style, i believe some indentation may be appropriate here :)

src/StatusBar.ts
```diff
@@ -41,32 +77,156 @@ export class StatusBar {
...
+
+    const messages = this.queue.map(item => {
+      const details = item.details ? `: ${item.details}` : ''
+      return `${item.source}: ${item.status} ${details}`
+    })
+    this.workspaceOutput.append(messages.join('\n'))
   }
+
+  private isMultiroot() {
+    return this.queue.length > 1
```
escaton: this is implementation detail, it needs to be more explicit

src/StatusBar.ts
```diff
@@ -41,32 +77,156 @@ export class StatusBar {
...
     this.queue = this.queue.filter(x => x.source !== source)
-    this.queue.unshift({
+    const item: QueueItem = {
       source,
       status,
       details,
-    })
-    this.updateStatus()
+    }
+    this.queue.unshift(item)
```
escaton: i just realize, that there is a memory leakage here. We should filter out all previous items with same source. Would you mind to add it?

src/StatusBar.ts
connectdotz: I agree. Actually, I think there are simply too many 'get x by statusType' kind of logic throughout the code, time to do some refactoring to consolidate those into a self-contained block...

src/StatusBar.ts
connectdotz: good catch

src/StatusBar.ts
connectdotz: even with indentation, I agree it could still be hard to read, sigh, breaking it up...

src/StatusBar.ts
connectdotz: not sure what you mean here... I wanted to check if there are multiple sources to decide if to show the summary status. It might make more sense after switching queue to Map (see below)

src/StatusBar.ts
connectdotz: I think we can be more efficient with a Map, not array/queue here. Since we really just need 1 status per source and the sequence is not important since you used the status priority as the determining factor... take a look of the updated StatusBar and see if it makes sense...

tests/StatusBar.test.ts
```diff
@@ -29,10 +33,31 @@ describe('StatusBar', () => {
...
+      registerCommand.mock.calls.forEach(c => {
+        // tslint:disable no-bitwise
+        if (c[0].includes('show-summary-output')) {
+          found |= 0x1
+        }
+        if (c[0].includes('show-active-output')) {
+          found |= 0x2
+        }
+      })
+      expect(found).toEqual(3)
```
stephtr: It is more performant, but is it also easier to read than something like:
```
const calls = registerCommand.mock.calls
expect(calls.some(c => c[0].includes('show-summary-output'))).toBe(true)
expect(calls.some(c => c[0].includes('show-active-output'))).toBe(true)
```
?

tests/StatusBar.test.ts
```diff
@@ -42,26 +67,12 @@ describe('StatusBar', () => {
       const helpers = statusBar.bind(source)
       ;['initial', 'running', 'success', 'failed', 'stopped'].forEach(status => {
         helpers[status]()
-        expect((statusBar as any).queue).toContainEqual({ source, status })
+        expect(updateStatusSpy).toHaveBeenCalledWith({ source, status })
```
stephtr: Are you sure that this is working? `{ source, status }` creates a new object and I thought that `toHaveBeenCalledWith` only checks identity of the reference (in contrast to `toContainEqual`)

src/StatusBar.ts
```diff
@@ -1,72 +1,254 @@
...
+  clearSpinner(): void
+  startSpinner(callback: () => void): void
+}
+
+const createStatusBarItem = (type: StatusType, priority: number): SpinnableStatusBarItem => {
+  let spinner: NodeJS.Timer | undefined
+  const item = vscode.window.createStatusBarItem(vscode.StatusBarAlignment.Left, priority)
+  const clearSpinner = () => {
+    if (spinner) {
+      clearInterval(spinner)
```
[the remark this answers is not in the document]
connectdotz: in statusBar `render()` , it did call the `clearSpinner()` so basically each `startSpinner` only call back for 1 interval unless it is started again within `render()`. Does that address your conern?

src/TestResults/TestResultProvider.ts
```diff
@@ -31,26 +42,65 @@ export class TestResultProvider {
   }
 
   getResults(filePath: string): TestResult[] {
+    const maybeTemplateLiteral = (s: string) => s.indexOf('$') > -1
```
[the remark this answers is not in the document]
connectdotz: that's a good suggestion, done

src/TestResults/TestResultProvider.ts
```diff
@@ -31,26 +42,65 @@ export class TestResultProvider {
...
+              err = 'failed to match test result, might be caused by template-literal test name?'
+            }
+          }
+          break
+        default:
+          // multiple matches, select according to the following criteria
+          assertion = findAssertionByLocation(test, match)
+          if (!assertion) {
+            // can't find the match, it could due to sourceMap related issue, let's try our best to locate one then
+            assertion = match.find(a => a.status !== TestReconciliationState.KnownFail) || match[0]
```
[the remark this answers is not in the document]
connectdotz: hmm... good question, I don't know the answer either... looks like it was added in this [commit](https://github.com/jest-community/vscode-jest/commit/4a5d225a31930b091e1fa0a0be85414bf66e2b82#diff-785a64be7997ebe2d7a9fdf2b7231468), maybe @ThomasRooney can shed some light?

src/StatusBar.ts
```diff
@@ -1,72 +1,254 @@
...
+    statusBarItem.clearSpinner()
+
+    const message = this.getMessageByStatus(request.status)
+    if (request.status === 'running') {
+      statusBarItem.startSpinner(() => this.render(request, statusBarItem))
+    }
+
+    switch (statusBarItem.type) {
+      case StatusType.active:
+        const details = !this.needsSummaryStatus() && request.details ? request.details : ''
```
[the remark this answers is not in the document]
connectdotz: I didn't see any lint rules/best-practice regarding the condition in ternary expression, however, did find the [multiline-ternary](https://eslint.org/docs/rules/multiline-ternary) in eslint, which defaults to separate lines. I could change it accordingly, and it will probably make the condition more readable.

src/extensionManager.ts
```diff
@@ -6,13 +6,15 @@ import { JestExt } from './JestExt'
 import { DebugCodeLensProvider, TestState } from './DebugCodeLens'
 import { DebugConfigurationProvider } from './DebugConfigurationProvider'
 import { IPluginResourceSettings, IPluginWindowSettings } from './Settings'
+import { statusBar } from './StatusBar'
 
 export class ExtensionManager {
+  debugCodeLensProvider: DebugCodeLensProvider
+  debugConfigurationProvider: DebugConfigurationProvider
+
   private extByWorkspace: { [key: string]: JestExt } = {}
```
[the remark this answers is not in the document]
connectdotz: from the assignment code in `register()`:
```
this.extByWorkspace[workspaceFolder.name] = new JestExt(
```
it should always be `JestExt`... not all workspace will be in the "extByWorkspace" dictionary, i.e. not all worksapces will be registered.

src/extensionManager.ts
```diff
@@ -79,15 +82,18 @@ export class ExtensionManager {
...
     return (
       // start if enabledWorkspaceFolders contains current workspace
       enabledWorkspaceFolders.includes(workspaceFolderName) ||
-      // dont start if disabledWorkspaceFolders doesn't contain current workspace
+      // don't start if disabledWorkspaceFolders doesn't contain current workspace
       !disabledWorkspaceFolders.includes(workspaceFolderName)
     )
   }
-  getByName(workspaceFolderName: string) {
+  public getByName = (workspaceFolderName: string) => {
```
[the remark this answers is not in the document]
connectdotz: arrow-function can capture "this" so it is easier to pass alone as a function-parameter, such as in the `extension.activate()`:
```
context.subscriptions.push(...statusBar.register(extensionManager.getByName),
```
instead of:
```
context.subscriptions.push(...statusBar.register((workspaceFolderName: string) => extensionManager.getByName),
```

tests/StatusBar.test.ts
connectdotz: I like that, especially in tests I agree we should strive for clarity than performance... done

tests/StatusBar.test.ts
connectdotz: as far as jest 24.7.0 goes, yes it worked, otherwise, the test will fail 😏

src/StatusBar.ts
connectdotz: @escaton did you get the chance to look at the updated file? does that address your request?

connectdotz:
ok, we have released the `jest-editor-support` and now this PR is finally able to be compiled and passed ci tests. Thanks for @stephtr and @escaton to review it, please take a look at the latest commit to see if you have any concern that hasn't been addressed yet.

src/StatusBar.ts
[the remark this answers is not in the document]
stephtr: My idea was to look at the statusBarItem as a separate object. If for example the implementation of `StatusBar` would some when change and by chance call `startSpinner` twice, the item would spin indefinitely. Therefore I would check within `startSpinner` that `spinner` is `undefined`. But that's just my personal opinion and to be honest completely unimportant.

src/StatusBar.ts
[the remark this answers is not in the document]
stephtr: `const details = (!this.needsSummaryStatus() && request.details) || ''` probably wouldn't be obvious enough?
Otherwise I'm also fine with leaving it as it is.

src/extensionManager.ts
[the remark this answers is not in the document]
stephtr: I was thinking about `extByWorkspace['not_existing_workspace'], but that was my mistake.

src/extensionManager.ts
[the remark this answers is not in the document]
stephtr: Ah ok, I didn't notice that.
(Somehow I still slightly prefer your second option or `.register(extensionManager.getByName.bind(extensionManager))` over changing an object method to an error function just that is easier to call it, but that was just my personal stubborn opinion.)

src/Coverage/CoverageCodeLensProvider.ts
```diff
@@ -2,15 +2,18 @@ import * as vscode from 'vscode'
...
 
-  constructor(getJestExt: (uri: vscode.Uri) => JestExt) {
+  constructor(getJestExt: GetJestExt) {
     this.getJestExt = getJestExt
   }
 
   public provideCodeLenses(document: vscode.TextDocument, _token: vscode.CancellationToken) {
-    const coverage = this.getJestExt(document.uri).coverageMapProvider.getFileCoverage(document.fileName)
+    const ext = this.getJestExt(document.uri)
+    const coverage = ext ? ext.coverageMapProvider.getFileCoverage(document.fileName) : undefined
```
stephtr: Since we return if `coverage` is falsy, what about `const coverage = ext && ext.coverageMapProvider.getFileCoverage(document.fileName)` instead?

src/DebugCodeLens/DebugCodeLensProvider.ts
```diff
@@ -7,12 +7,13 @@ import { TestReconciliationState } from '../TestResults'
 import { TestState, TestStateByTestReconciliationState } from './TestState'
 import { JestExt } from '../JestExt'
 
+type GetJestExt = (uri: vscode.Uri) => JestExt | undefined
```
stephtr: Since it's the second time defined that way, would it be worth the effort to outsource it (for example to `vscode-jest/JestExt.ts`)?

src/Coverage/CoverageCodeLensProvider.ts
connectdotz: great idea!

src/DebugCodeLens/DebugCodeLensProvider.ts
connectdotz: we could,  but it is not a method defined in `JextExt` but from `extensionManager`, so I move the type alias there instead. 

The real concern here is the fact that typescript didn't catch the original `undefined` type error and caused run time crash so I refactor the code and introduced an explicit typedef, in my opinion, a bit unnecessary if tsc does its job right...

src/StatusBar.ts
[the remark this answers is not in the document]
connectdotz: i see what you mean, and I think you are right, it is definitely more defensible for future development, in case spinner did get misused... added `clearSpinner` at the beginning of each `startSpinner` should do the trick.

connectdotz:
sorry, was busy in the last few days, I need to re-merge with the latest beta changes and then we should be good to go...

stephtr:
To me it seems like [your merge](https://github.com/jest-community/vscode-jest/pull/447/commits/74fcb6605fab50ab615356b78d9e1848663ccf42) went wrong, undoing previous changes.

src/TestResults/TestResultProvider.ts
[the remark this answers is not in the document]
connectdotz: I think I figure this one out, I think it was assuming the "failed" test will have the line number matched already, so it is only focused on matching the non-failed ones, but of course, this assumption might not always be true since any location matching could fail due to transpile/sourcemap... 

while adding more tests to this, I discover an edge case for duplicate test names. Also didn't feel comfortable about template-literal use case, which right now will fall back to location matching, kind of dangerous as mentioned above. Added partial name match to increase correctness, when in doubt (such as multiple partial matched names found), we will just mark it as 'unknown' and log the info in console (in debug mode).

@stephtr if you get a chance, please take a look to make sure I didn't miss anything obvious, appreciate it.

connectdotz:
ci failed due to test coverage rounding, anyway, I decided to beef up the tests coverage (we are +2.2% now 😄).  Plus a bonus: discovered a new bug in TestResultProvider, good karma 🙏 ...

PR #459, opened 2019-05-24
Add disableOptimisticBPs option to debug configuration

### LINKED ISSUE #400: Cannot hit breakpoint inside a test with VS Code IDE
### Environment

1. `node -v`: [v10.12.0]
1. `npm -v`: [6.4.1]
1. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): [jest@23.6.0]
1. your vscode-jest settings if customized:
    - jest.pathToJest? [not set]
    - jest.pathToConfig? [not set]
    - anything else that you think might be relevant? []

1. Operating system: [Ubuntu 18.04]

### Prerequisite
- are you able to run jest test from command line? [yes]
- how do yo run your tests from command line? (for example: `npm run test` or `node_modules/.bin/jest`) [npm run test]

### Steps to Reproduce

Here is a really simple repository that show it is not possible to hit breakpoint into the file user.repository.spec.ts on line 12 const users = await userRepository.getAll(); 
https://github.com/stherrienaspnet/vscode-jest-debug.git

### Relevant Debug Info

[Did not found any resolution yet]

### Expected Behavior

[Should break on line 12 when there is a breakpoint and press play vscode-jest-test in launch menu ]

### Actual Behavior

[Break on line 12 is not hit]

Thanks you very much!
--- 

The fastest (and the most fun) way to resolve the issue is to submit a pull-request yourself. If you are interested, feel free to check out the [contribution guide](https://github.com/jest-community/vscode-jest/CONTRIBUTING.md), we look forward to seeing your PR...

blikblum:
Fixes #400 

See full context and reasoning at https://github.com/microsoft/vscode/issues/60187

Also note that in [VS Code recipe](https://github.com/Microsoft/vscode-recipes/tree/master/debugging-jest-tests#configure-launchjson-file-for-your-test-framework), which is used as [reference](https://github.com/jest-community/vscode-jest/blob/master/src/DebugConfigurationProvider.ts#L50) to the debug configuration, was updated to add disableOptimisticBPs option


src/DebugConfigurationProvider.ts
```diff
@@ -60,6 +60,7 @@ export class DebugConfigurationProvider implements vscode.DebugConfigurationProv
       cwd: '${workspaceFolder}',
       console: 'integratedTerminal',
       internalConsoleOptions: 'neverOpen',
+      disableOptimisticBPs: true,
```
connectdotz: my gut feeling is we probably don't want to add this for **all** debug config... shouldn't we just add this to the CRA [snippet](https://github.com/jest-community/vscode-jest/blob/6718f847a90e6fcae1faf847db289e456190b4b4/package.json#L167) in project.json instead? 

@stephtr probably is most familiar with this...

blikblum: > add this to the CRA snippet in project.json 

This is not specific to CRA. I use it for non React projects. 

And as i pointed is the recommended setting by Microsoft to run Jest: https://github.com/Microsoft/vscode-recipes/tree/master/debugging-jest-tests#configure-launchjson-file-for-your-test-framework

blikblum: Finally, in microsoft/vscode#60187 has all the reasoning as this is necessary to run jest properly specially https://github.com/microsoft/vscode/issues/60187#issuecomment-428408469

connectdotz: looks like I was able to debug typescript projects without this flag, but js project requiring transpiling does benefit from having this flag. Just to play devil's advocate, if there is no downside of enabling this flag, why not use it for all the other debug recipes?

blikblum: > looks like I was able to debug typescript projects without this flag

typescript is not affected

>  if there is no downside of enabling this flag, why not use it for all the other debug recipes?

Seems its not as performant as disabled because needs to wait to the sourcemap be loaded to set the breakpoint. Although i did not notice any slowdown

Anyway, another possibility is to add an option disabled by default. If you can point me the place where to add i would be glad

stephtr: >> if there is no downside of enabling this flag, why not use it for all the other debug recipes?
>
> Seems its not as performant as disabled because needs to wait to the sourcemap be loaded to set the breakpoint. Although i did not notice any slowdown

According to the VS Code issue it can be (probably mostly for larger projects), that the breakpoints aren't ready when the tests are being executed. In that case one has to run the tests a second time or add a `debugger` statement to the top of the file and wait for a bit. The `disableOptimisticBPs` fix seems to be just kind of temporarily ("It's not a permanent solution but should unblock this at least."), but as long as it is used in vscode-recipes, I would set it by default in vscode-jest.

stephtr: By the way, if you already update this file, could you update the link on line 53 to [facebook.github.io](https://facebook.github.io/create-react-app/docs/debugging-tests#debugging-tests-in-visual-studio-code)?

stephtr:
Thanks for this PR! Additionally you should also update the templates within `package.json`, then I would be fine with it.

src/DebugConfigurationProvider.ts
blikblum: > According to the VS Code issue it can be (probably mostly for larger projects)

I get this issue in small projects like https://github.com/blikblum/nextbone-state/blob/master/test/unit/resource.spec.js

> In that case one has to run the tests a second time or add a debugger statement to the top of the file and wait for a bit

I found this by myself. Before this option, I put an aleatory breakpoint at top of file and when is hit i put in the correct place. 
With this option i not needed to do it anymore.

blikblum:
> Thanks for this PR! Additionally you should also update the templates within package.json, then I would be fine with it.

Done

> By the way, if you already update this file, could you update the link on line 53 to facebook.github.io?

Done

connectdotz:
I have only 1 concern: will this flag negatively impact the use cases that work fine now (without this flag)? 

The thinking goes like this:
- There is probably some advantage for vscode to preload the sourcemap, otherwise, they can just disable the preload all together rather than using this flag to workaround... 
- while it is definitely a great improvement for use cases that used to be broken, but by adding this flag to **all** use cases, are we fixing the ones that **don't need to be fixed** and degrading their user experience as a result? 

I don't know how big or small the experience degradation will be... If you are pretty confident such degradation is not noticeable for most cases then all is fine; if you are not quite sure either (and do not want to consider _debug config customization_), then maybe rebase it to `beta` so we can test this a bit more before releasing to all users?

blikblum:
Looking at https://github.com/microsoft/vscode-node-debug2/commit/c9feb129aa3667b8e89ba60423265c645bc95522#diff-1b67da1705d67aa2b3d4a722dab1c0f4R21 comments

Makes believe that for not transpiled (no source map) code it will not set breakpoint

Maybe @roblourens  can answer this question and the performance impact of such option

blikblum:
> Makes believe that for not transpiled (no source map) code it will not set breakpoint

I created a sample project without any kind of transform, and breakpoints are working fine with or without disableOptimisticBPs option

roblourens:
I believe the actual result will be that breakpoints are bound when a sourcemap is loaded that covers that scripts OR when the script itself is loaded (if it is not transpiled).

But there is a potential impact to projects that don't use sourcemaps. Since the breakpoint would be bound until the script is actually loaded (and already running) it introduces a race condition where a breakpoint in a require()d script may not be hit if the code on that line runs before the breakpoint can be bound. So I think it's better if this flag is only used on projects that have sourcemaps. The debug adapter cannot tell the difference because it doesn't know whether breakpoints are involved until the runtime script is loaded.

stephtr:
Even if it negatively affects some users (breakpoints working only at the second time or with an additional debugger statement), I would go for the change, since it is [recommended](https://github.com/Microsoft/vscode-recipes/tree/master/debugging-jest-tests#configure-launchjson-file-for-your-test-framework) by the VS Code maintainers. We can add a small note to our readme file, that that part of the debug configuration can be customized, though.

blikblum:
I can add the note but i have a question before doing it:

The configuration templates can be configured by editing it in launch.json but the configuration that is used when the debug is started by codelens can be configured?

connectdotz:
@blikblum 

> The configuration templates can be configured by editing it in launch.json but the configuration that is used when the debug is started by codelens can be configured?

the codelens will use the launch.json config if it exists, the hardcoded one in code is only the **fallback** when we can't find the existing debug config.

@stephtr 

> Even if it negatively affects some users ... I would go for the change, since it is recommended by the VS Code maintainers.

looks like the flag is added to vscode-recipes [in this commit](https://github.com/microsoft/vscode-recipes/commit/ad504728327c83aa88fe78381fea273971d4ad68) by @roblourens, who have just shared with us his thought: 

> So I think it's better if this flag is only used on projects that have sourcemaps

---
Given we also don't know which project requires transpiling/sourcemaps, and many projects are working fine without the flag, I am not sure **'adding it for all projects'** is the only way forward... I feel we should at least have a discussion of the alternative: 
- adding the flag to CRA snippet because react will definitely transpile files
- and document this flag in the debug troubleshooting section, as one of the potential solutions people can try if their BP looked wrong and how they can customize their launch.json... (there might be other reasons why BP won't work even with this flag, such as if the sourucemap is not generated; would be valuable if we add those as well).
    
there might be other alternatives, in general, my 2 cents:  it is probably better to only fix the ones that are broken and avoid introducing unnecessary changes to break the ones that are working (regression).  Unless -- we are sure the race condition happens very very rarely, and the majority of our users will need this flag -- then it does make sense to add it for everybody.

blikblum:
I'm fine with either approach. Let me know to update the PR

connectdotz:
@stephtr what do you think?

stephtr:
> looks like the flag is added to vscode-recipes in this commit by @roblourens, who have just shared with us his thought

Ah thanks, I didn't get that connection. @roblourens wouldn't it be then better for vscode-recipies to remove the `disableOptimisticBPs` property and add a note instead that one should add it when using sourcemaps?

Even though I would have preferred sticking to vscode-recipies, @connectdotz proposal is probably the best way to go.

roblourens:
I'm not sure what vscode-recipes should do, it is a stripped down example but I would think that most real projects are using some sort of transpilation and sourcemaps, and it's easier to just include the option rather than try to explain what it does and when it should be used.

connectdotz:
@roblourens you have a point, but to look at it from another angle: when the projects, such as typescript projects (a rapidly growing community), start to experience race condition you mentioned, which they didn't have before, wouldn't we need to explain to them why they are having this problem and instruct them how to remove the flag? So it seems to me, adding the flag or not, we can't really avoid the education part...

IMHO, to educate people on how to fix their existing problem improves their user experience; on the other hand, explain to people why they have to deal with a new regression problem after upgrade decreases user experience... If all things are equal, one would think the former is preferred.

To add another consideration, you introduced 'disableOptimisticBPs' and not _'enableOptimisticBPs'_, why? You want the optimal performance, only when it can't be achieved then users can add the 'disableOptimisticBPs', i.e. when the tool can't predetermine, it defaults to the preferred, more optimal solution. I think `vscode-jest` could benefit from the same thinking...

connectdotz:
@roblourens actually thinking further, I think we might have an opportunity to resolve the root cause and deliver the optimal debug experience... want to verify the following:

- in microsoft/vscode#60187, you seem to indicate that the problem is that vscode doesn't know where the sourcemaps are, otherwise one can just use the `outFile`/`sourceMap` in Launch.json to help vscode resolve BP, right? 

- back to our extension `vscode-jest`: it, among other features, provides debug codelens/snippets for jest tests, **after** running the whole jest tests suite. Therefore we should be able to provide the sourcemaps from jest: if we run jest with specific `cacheDirectory` then pass that to `outFile` in Launch.json, will that work?

- question for jest, @SimenB does jest reset cacheDirectory? we usually run all tests upon extension launch, can we expect the sourcemaps for all tests are available, i.e. not being reset, for subsequent runs?

roblourens:
TS projects aren't affected here, what I call "optimistic bps" are only for setting the breakpoint in a .js file. TS (when the sourcemaps are not preloaded) will always have the race condition issue. As for which should be the default, I think having it enabled by default is the right course, because it should only be disabled for projects with JS sourcemaps that can't be preloaded, which I consider the edge case. Plus some of these projects that are built such that line numbers are preserved between the original and transpiled outputs, and those will be fine without the flag. That is why this came up originally, I think in create-react-app, when they disabled the babel flag that preserves line numbers. JS projects without sourcemaps are probably more common so I optimize for that.

Using outFile is the preferred thing to do. When I looked at cacheDirectory, I think it didn't help but I can't remember exactly why. I think the problem was files are built on the fly so I can't preload from there before starting the program. But you might know better than me. If you load sourcemaps from a previous run then they will be inaccurate for files that have changed, and I don't think the debug adapter will replace the loaded sourcemap if it turns out to have been inaccurate.

Another thing to consider. The "real" solution to all of this, in my mind is having the ability to pause execution each time a script is loaded, giving the debugger a chance to determine whether that script has a sourcemap, load it, resolve BPs, then continue executing that script. I proposed this quite a while ago and the feature was recently merged into V8 - https://chromium-review.googlesource.com/c/v8/v8/+/1354245 so, we will have a better solution in some future node version.

SimenB:
>does jest reset cacheDirectory?

no, not unless passed `--clear-cache` or `--no-cache` (the latter _does_ cache things, it just clears it first)

>we usually run all tests upon extension launch, can we expect the sourcemaps for all tests are available, i.e. not being reset, for subsequent runs?

Yes. But they might not be accurate.
- files change, and we haven't written the new sourcemap to disk yet
- toggling on coverage or otherwise changing babel's (or w/e transformer you use) config

---

(answering in isolation, haven't read this thread)

connectdotz:
> Yes. But they might not be accurate.

I should mention `vscode-jest` always has a background jest in watch mode, we update the test status and the debug codelens whenever the file changes with jest test result, so I would think the cache should be pretty accurate... However, I can't find sourcemaps or the generated code inside the cache directory, where should I look?

SimenB:
Do `jest --show-config`, then look at `cacheDirectory`. Inside of that directory there'll be a `jest-transform-cache-*somehash*` directory which contains the transformed files and any source maps.

The logic for this lives here: https://github.com/facebook/jest/blob/b4e07d485295946f4dd4d5683d3d74ca88bf48dc/packages/jest-transform/src/ScriptTransformer.ts#L111-L134

connectdotz:
ok, I see what the problem is now: even though we can set the `cacheDirectory`, jest is not emitting the transpiled files (names) in the way the debugger can use... 

I start to see @roblourens point now, indeed trying to preload sourcemap from disk could be complicated, while more ducktapes can be applied, I agree a better and more robust solution is as you mentioned:

> The "real" solution to all of this, in my mind is having the ability to pause execution each time a script is loaded

ok, so let me see if I can get this right (please correct, if not): 
- seems like this flag only apply to debugging **js** files
- who will be negatively impacted? the js files with sourcemaps on disk that used to be preloaded, now with this flag, they will be subject to potential race condition. 
- ts files without sourcemaps on disk have always been non-optimistic, i.e. subject to the race condition. (I assume if they have sourcemaps on disk, even with this flag, the debugger will still try to preload?)

@roblourens, in your experience, would you say the race condition is relatively rare? As I don't recall flaky BPs for our ts projects (without sourcemap on disk)... 

if a race condition is far and between, then I think it is ok to apply it indiscriminately here (we should still document it)... once the nodejs patch goes live, the _rare_ race condition should be further reduced/eliminated.

roblourens:
>  the js files with sourcemaps on disk that used to be preloaded, now with this flag, they will be subject to potential race condition.

No, if sourcemaps **are** on disk and preloaded, they will still be preloaded and that case is not affected by this flag. The rest you say is correct.

> @roblourens, in your experience, would you say the race condition is relatively rare?

It entirely depends on the the user's machine and random chance. If the breakpoint is right at the top of the file, quite likely. If the breakpoint won't be hit for a bit - e.g. there are other tests that run first, less likely.

connectdotz:
> No, if sourcemaps are on disk and preloaded, they will still be preloaded and that case is not affected by this flag.

ok, so the only scenario that could see negative impact is the js files without sourcemap, now they might see race condition, any other use case you can think of?

connectdotz:
@stephtr @blikblum after these conversations, I feel a lot more comfortable about this change, sorry took a while to reach this point. If you guys want to release it in the next version, you can rebase it to `beta`, I will try to cut a pre-release this week...

blikblum:
> you can rebase it to beta

Done

SimenB:
>even though we can set the `cacheDirectory`, jest is not emitting the transpiled files (names) in the way the debugger can use...

It's not currently designed for consumption by others than jest - but that is something we can potentially change, especially if it improves the debug story in IDEs. That might potentially take some time though, so probably a good idea to land this PR regardless

connectdotz:
merged into beta, thanks @blikblum and @stephtr

PR #461, opened 2019-06-03
Restart command

escaton:
This is a copy of https://github.com/jest-community/vscode-jest/pull/441 with one extra commit to fix tests

### DIFF
--- a/src/JestProcessManagement/JestProcess.ts
+++ b/src/JestProcessManagement/JestProcess.ts
@@ -9,7 +9,7 @@ export class JestProcess {
   private projectWorkspace: ProjectWorkspace
   private onExitCallback: Function
   private jestSupportEvents: Map<string, (...args: any[]) => void>
-  private resolve: Function
+  private resolve: Function | null
   private keepAliveCounter: number
   public keepAlive: boolean
   public stopRequested: boolean
@@ -17,6 +17,7 @@ export class JestProcess {
 
   private startRunner() {
     this.stopRequested = false
+    this.resolve = null
     let exited = false
 
     const options = {
@@ -39,7 +40,7 @@ export class JestProcess {
           if (this.onExitCallback) {
             this.onExitCallback(this)
           }
-          if (this.stopRequested) {
+          if (this.resolve) {
             this.resolve()
             this.resolve = null
           }
--- a/src/JestProcessManagement/JestProcess.ts
+++ b/src/JestProcessManagement/JestProcess.ts
@@ -72,13 +72,13 @@ export class JestProcess {
     this.onExitCallback = callback
   }
 
-  public onJestEditorSupportEvent(event, callback) {
+  public onJestEditorSupportEvent(event: string, callback: (...args: any) => void) {
     this.jestSupportEvents.set(event, callback)
     this.runner.on(event, callback)
     return this
   }
 
-  public stop() {
+  public stop(): Promise<undefined> {
     this.stopRequested = true
     this.keepAliveCounter = 1
     this.jestSupportEvents.clear()
@@ -88,7 +88,7 @@ export class JestProcess {
     })
   }
 
-  public runJestWithUpdateForSnapshots(callback) {
+  public runJestWithUpdateForSnapshots(callback: Function) {
     this.runner.runJestWithUpdateForSnapshots(callback)
   }
 }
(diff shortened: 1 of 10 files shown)


src/JestProcessManagement/JestProcess.ts
```diff
@@ -35,8 +35,10 @@ export class JestProcess {
         if (--this.keepAliveCounter > 0) {
           this.runner.removeAllListeners()
           this.startRunner()
-        } else if (this.onExitCallback) {
-          this.onExitCallback(this)
+        } else {
```
escaton: @connectdotz could you check if this condition is now done right?
Looks like without this change `resolve` was never called until `onExitCallback` was also set.

escaton:
Btw, if you take a look at [old coverage](https://coveralls.io/builds/23529501/source?filename=src%2FJestProcessManagement%2FJestProcess.ts#L65) (before my commit) you'll see something weird
изображение
while in fact uncovered lines were here

изображение
That is a sourcemap issue which i once tried to fix in https://github.com/jest-community/vscode-jest/pull/381
It's a bit obsolete now, i'll update it.

src/JestProcessManagement/JestProcess.ts
connectdotz: good catch 👍

src/JestProcessManagement/JestProcess.ts
```diff
@@ -35,8 +35,10 @@ export class JestProcess {
...
+          if (this.onExitCallback) {
+            this.onExitCallback(this)
+          }
           if (this.stopRequested) {
```
connectdotz: now that we do set this.resolve to null, I think we should check `this.resolve` instead of `this.stopRequested` here. As long as there is a`this.resolve`, we should just resolve it.

vdh:
Everything's looking good from what I can tell 👍

src/JestProcessManagement/JestProcess.ts
escaton: Okay, checking here for `this.resolve` which i rename to `this.stopResolveCallback` (maybe better naming?) and turning `this.stopRequested` into method.

src/JestProcessManagement/JestProcess.ts
```diff
@@ -85,16 +83,15 @@ export class JestProcess {
...
       this.jestSupportEvents.clear()
       this.runner.closeProcess()
 
       // As a safety fallback to prevent the stop from hanging, resolve after a timeout
       // TODO: If `closeProcess` can be guarenteed to always resolve, remove this
       setTimeout(() => {
-        if (this.resolve === resolve) {
+        if (this.stopResolveCallback === resolve) {
           resolve()
         }
```
connectdotz: just notice this, what happens when they are not the same? is there any reason we should not resolve() regardless? I would also imagine the way to trigger this race condition is to issue 2nd `stop()` before the first one completed: the 2nd stop() will override the `this.stopResolveCallback` (how about calling it "processStopped"?), we should have a test and handle this scenario, otherwise, I suspect the first stop() will probably hang, even with the timeout...

vdh: Ah yes, that would cause a hang. It should be safe to remove the conditional and simplify it to just `resolve()` to avoid the hang, since multiple `resolve` calls are no-ops.

Ideally (perhaps in a future change), it would be good to refactor it away from creating the promise in `stop`, and create and store the promise inside the `startRunner` call instead. That would eliminate that ugly `resolve` ivar. Probably pair it with the `runner` ivar, as something like an `onExit` promise ivar. That could also mean wrapping/replacing/deprecating the `onExitCallback` API at the same time to keep things consistent.

escaton: ok, i removed condition and added test case

tests/JestProcessManagement/JestProcess.test.ts
```diff
@@ -216,6 +216,14 @@ describe('JestProcess', () => {
       await promise
       expect(promise).resolves.toBeUndefined()
     })
+
+    it('do not hangs on multiple stop() calls', async () => {
+      const promise = jestProcess.stop()
```
connectdotz: maybe add `expect.assertions(1)` to ensure test will be completed?

PR #476, opened 2019-07-03
Fix security Alerts, remove `create-react-app` example

stephtr:
- For `webpack-dev-server` and `merge`: I removed the `create-react-app` example. If one would like to have it locally available, just run `npx create-react-app`.
~I updated the whole `create-react-app` template/example to the newest version. @connectdotz: Since this repo contains just a plain `create-react-app` bootstrapped app, I would be in favour of completely removing the integrations folder.~
- `randomatic`: I got rid of that by removing the unused `micromatch` dependency.
- `cryptiles`: It was enough to trigger an update of the secondary dependency.


connectdotz:
> Since this repo contains just a plain create-react-app bootstrapped app, I would be in favour of completely removing the integrations folder.

I agree.

Issue #493, opened 2019-08-21
ansi code escape from jest output with @testing-library/react


MrLuje:
### Environment

1. `node -v`: v10.12.0
1. `npm -v`: 6.4.1
1. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): jest@24.8.0
1. your vscode-jest settings if customized:
    - jest.pathToJest? node_modules/.bin/jest --no-color

1. Operating system: windows 10

### Prerequisite
- are you able to run jest test from command line? yup
- how do you run your tests from command line? `npm run test`

### Steps to Reproduce

There is a known issue with jest ouput in tooltip if colors are enabled which can be solved with the --no-color arg (cf #279)

--no-color fixes this issue but it doesn't seem to impact colors from [@testing-library/react](https://github.com/testing-library/react-testing-library) which colorizes DOM elements on failing tests

Not sure if the issue can be solved from Jest's side only, though

![image](https://user-images.githubusercontent.com/632075/63422458-b1032080-c40a-11e9-970c-314766742979.png)

![image](https://user-images.githubusercontent.com/632075/63423207-1572af80-c40c-11e9-85c4-64f7fc86af2c.png)

Issue #279, opened 2018-03-06
ansi code escape from jest output


tardyp:
Using standard ```vue-cli init webpack ```

upon test errors, I got problems and inline error messages corrupted with ansi codes:

![image](https://user-images.githubusercontent.com/109859/37044926-06d495f6-2165-11e8-801c-8d28013efd7c.png)

I can hack them out using this in workspace config:

    "jest.pathToJest": "node_modules/.bin/jest --no-color",

but in that case the 'debug' button will fail saying it is not able to find jest.js:
![image](https://user-images.githubusercontent.com/109859/37045095-6f834cfa-2165-11e8-8453-6036a4634432.png)

seanpoulter:
That's a bit strange. Would you have time to debug the extension and see where we've got the bonus ANSI? The output is coming from Jest as JSON using the `--json` argument. If you run Jest from the command line with that, are the ANSI codes in there as well?

PR #271 is in the works and should help with the debugger finding `jest.js`.

tardyp:
@seanpoulter
Sorry for delay.
 I ran jest with --json and it includes ansi codes

grahamb:
I'm seeing this as well with the latest plugin on VS Code 1.21.0. Running manually with `--json` includes ANSI escape codes in the output.

![screenshot 2018-03-20 11 43 25](https://user-images.githubusercontent.com/146111/37675992-f29d5b16-2c33-11e8-9236-fcb7da13fd4b.png)

JSON output:
```json
{
	"numFailedTestSuites": 1,
	"numFailedTests": 1,
	"numPassedTestSuites": 0,
	"numPassedTests": 1,
	"numPendingTestSuites": 0,
	"numPendingTests": 0,
	"numRuntimeErrorTestSuites": 0,
	"numTotalTestSuites": 1,
	"numTotalTests": 2,
	"snapshot": {
		"added": 0,
		"didUpdate": false,
		"failure": false,
		"filesAdded": 0,
		"filesRemoved": 0,
		"filesUnmatched": 0,
		"filesUpdated": 0,
		"matched": 0,
		"total": 0,
		"unchecked": 0,
		"uncheckedKeys": [],
		"unmatched": 0,
		"updated": 0
	},
	"startTime": 1521571392982,
	"success": false,
	"testResults": [{
		"assertionResults": [{
			"ancestorTitles": [],
			"failureMessages": [],
			"fullName": "adds 1 + 2 to equal 3",
			"location": null,
[...]

PR #501, opened 2019-09-22
Remove ANSI color characters from diagnostic messages

jmarceli:
This MR should solve https://github.com/jest-community/vscode-jest/issues/493 as well as https://github.com/jest-community/vscode-jest/issues/279.

Before this MR output looks like this:
vscode-jest-ansi-colors

and after:
vscode-jest-no-ansi

### DIFF
--- a/src/JestExt.ts
+++ b/src/JestExt.ts
@@ -11,7 +11,7 @@ import {
   resultsWithLowerCaseWindowsDriveLetters,
   SortedTestResults,
 } from './TestResults'
-import { pathToJest, pathToConfig } from './helpers'
+import { pathToJest, pathToConfig, cleanAnsi } from './helpers'
 import { CoverageMapProvider } from './Coverage'
 import { updateDiagnostics, updateCurrentDiagnostics, resetDiagnostics, failedSuiteCount } from './diagnostics'
 import { DebugCodeLensProvider } from './DebugCodeLens'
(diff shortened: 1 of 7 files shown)


jmarceli:
I've implemented ANSI characters removal like it was before, so with a regex, but maybe it would be better to use some external library e.g. https://www.npmjs.com/package/strip-ansi
Please let me know what is your opinion. I didn't want to introduce a new dependency that is why I choose to stay with the existing regex approach.

connectdotz:
> I didn't want to introduce a new dependency that is why I choose to stay with the existing regex approach.

I agree. Did you notice anything missing from the existing regex pattern?

src/JestExt.ts
```diff
@@ -459,7 +460,8 @@ export class JestExt {
   }
 
   private updateWithData(data: JestTotalResults) {
-    const normalizedData = resultsWithLowerCaseWindowsDriveLetters(data)
+    const noAnsiData = resultsWithoutAnsiCharacters(data)
```
connectdotz: a bit nitpicking here, technically we are just removing the ANSI escape sequence, not all ANSI characters... rename accordingly would make it less confusing.

jmarceli: Good point. Fixed.

jmarceli:
Regarding regex pattern, I don't know if there are any unhandled cases. I just prefer to use libraries instead of writing regexes if possible, but I think that we can leave it as it is for now.

Issue #503, opened 2019-10-01
Open failed snapshots in a diff view


borekb:
It would be awesome if I could open a failed snapshot expectation in a diff view. Right now, everything is condensed into a single line and using `jest` on the command line is my best bet (but the CLI isn't the best UX for this either).

borekb:
Related but (I think) different:

- Feature request - Formatted snapshot #387
- Snapshot previews #73

PR #510, opened 2019-10-12
Use short message and add test

jmarceli:
This PR replaces a terse message with a short one according to the https://github.com/jest-community/jest-editor-support/pull/18 discussion. It should improve `vscode-jest` plugin user experience as a terse message is often much less readable than short.
Simple example:

Terse:
terse-message

Short:
short-message

What is more, for specialized testing frameworks like [@testing-library/react](https://testing-library.com/docs/react-testing-library/intro) terse messages may be totally useless.

I hope that this PR will improve problems like this one https://github.com/jest-community/vscode-jest/issues/503, in a multiline view, snapshot diffs should be much more readable.


jmarceli:
I have no idea why `yarn danger` is failing and breaking CI build. Please let me know what I can do to fix that issue.

jmarceli:
Good catch. Please tell me if I understand your point correctly. I've removed `hover` key from a decorator generated by `generateInlineErrorDecorator`, so it should only show terseMessage for inline messages without overriding hover message which uses `vscode.Diagnostic`.

connectdotz:
I wasn't thinking about removing it, but I think what you did is equally fine.

PR #511, opened 2019-10-13
Change default value for inline error messages to disabled

jmarceli:
According to the https://github.com/jest-community/jest-editor-support/pull/18 discussion (view the latest comments), it might be a good idea to disable inline error messages by default as they are not very helpful and clutters the editor workspace (this just an opinion). What is more default Typescript related errors are also displayed only through underlining problematic phrase, without inline error message.
Feel free to criticize that idea :)


stephtr:
> However, if you guys all feel the same, we could try to set the jest.enableInlineErrorMessages default to false (in package.json) in the next major release and see how it goes...

> I am planning 3.1.0 release next week

@connectdotz Should we take that opportunity or wait?

connectdotz:
let's do it, @jmarceli if you can resolve the conflict I will merge it for 3.1

douglascayers:
Thanks for this change, just realized my Jest plugin got updated and I really enjoy the new default behavior. (I wasn't even aware I could change the behavior if that was a configurable option). This experience feels like Jest support is natively supported, I wouldn't have known I was using a plugin. This is a great add-on for Jest developers, well done and thank you :)

image

image

PR #513, opened 2019-10-13
Add Running tests message to StatusBar

jmarceli:
There is no information when tests are running in the StatusBar (besides initial tests run). This might be confusing because the user doesn't know if he should wait for tests to finish or maybe all tests are successfully passed. This commit adds the "Running tests" message after each file save.

There is also a minor improvement which is the capitalization of the "Initial full test run" message to stay consistent with the capitalization of status messages.


seanpoulter:
Thanks @jmarceli. Should we be checking if Jest is running before we change the message? 🤔

jmarceli:
Thanks for the comment. I've added a check for running Jest processes number.

src/JestExt.ts
```diff
@@ -301,6 +301,15 @@ export class JestExt {
     this.triggerUpdateActiveEditor(editor)
   }
 
+  onDidSaveTextDocument(_editor: vscode.TextDocument) {
+    if (!this.jestProcessManager.numberOfProcesses || this.jestProcessManager.numberOfProcesses < 1) {
+      return
+    }
+
+    // each file save will/should trigger tests re-running
+    this.status.running('Running tests')
```
connectdotz: the problem of doing it here is when users updated a document that didn't trigger jest run (such as a Readme.md), the status will be spinning forever... it's kind of hard to guess which file will trigger jest to re-run the tests since jest configuration could include custom matching logic/patterns...

jmarceli: I didn't think about that issue. I will try to find some reasonable solution, thanks for the comment.

connectdotz:
there is one thing we could try...  if we add a simple [jest reporter](https://jestjs.io/docs/en/configuration#reporters-array-modulename-modulename-options) to write out a simple text to stdout/stderr on `onRunStart`, vscode-jest listener can then parse and display the visual indicator. 

The reporter could live in `vscode-jest` but you will need to add an option to pass the [--reporters](https://jestjs.io/docs/en/cli#reporters) cli argument to `jest-editor-support`...

Is this something you would be interested to explore?

jmarceli:
@connectdotz thanks for the idea, I will try to implement that

jmarceli:
@connectdotz Please let me know what do you think about this solution

src/JestProcessManagement/JestProcess.ts
```diff
@@ -68,9 +68,12 @@ export class JestProcess {
     this.stopResolveCallback = null
     let exited = false
 
+    const reporterPath = `${process.cwd()}/src/JestProcessManagement/vscode-jest-reporter.ts`
```
connectdotz: this didn't seem to work:`process.cwd()` returns "/"

src/JestProcessManagement/JestProcess.ts
```diff
@@ -68,9 +68,12 @@ export class JestProcess {
...
     const options = {
       noColor: true,
       shell: platform() === 'win32',
+      // workaround required until reporters argument won't be supported by jest-editor-support
+      testFileNamePattern: `--reporters="${reporterPath}"`,
```
connectdotz: can jest actually run a typescript reporter? I suspect you will have to pass the js file instead...

also, I know this is just a temporary workaround, once the implementation worked, we should use a new option argument and patch `jest-editor-support` accordingly.

src/JestProcessManagement/JestProcess.ts
jmarceli: Do you have any idea how to reference extension directory in vscode? This works in my case probably because of running in development mode.

src/JestProcessManagement/JestProcess.ts
jmarceli: I'll try with that TS in .js idea. Pure .ts file didn't work. As for workaround I'll make PR to jest-editor-support when this PR will be ready.

src/JestProcessManagement/JestProcess.ts
connectdotz: [extensionPath](https://code.visualstudio.com/api/references/vscode-api#Extension) seems promising...?

jmarceli: Thanks, I've just used `extensionName` instead of `extensionId` which obviously doesn't work as expected.

src/JestProcessManagement/JestProcess.ts
jmarceli: It seems to be working with `.ts` reporter file.

src/JestExt.ts
```diff
@@ -412,16 +413,25 @@ export class JestExt {
     this.channel.appendLine(noANSI)
   }
 
+  private handleJestEditorSupportEvent(output: string) {
+    if (output.includes('onRunStart')) {
+      this.status.running('Running tests')
```
connectdotz: now we have a precise knowledge of jest starts, I think there is a great opportunity to consider killing the `clearOnNextInput`, which at times cleared the output channel prematurely. Do you think we can optimize and consolidate that logic based on 'onRunStart' instead?

jmarceli: Good insight. I removed `clearOnNextInput` as it looks to be not needed anymore.

src/JestProcessManagement/JestProcess.ts
```diff
@@ -68,9 +71,14 @@ export class JestProcess {
...
     let exited = false
 
+    const extensionPath = vscode.extensions.getExtension(extensionId).extensionPath
+    const reporterPath = join(extensionPath, 'out', 'reporter.js')
+
     const options = {
       noColor: true,
       shell: platform() === 'win32',
+      // workaround required until reporters argument won't be supported by jest-editor-support
+      testFileNamePattern: `--reporters="${reporterPath}"`,
```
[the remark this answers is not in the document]
jmarceli: Done

src/JestExt.ts
```diff
@@ -412,16 +402,27 @@ export class JestExt {
...
   }
 
+  private handleJestEditorSupportEvent(output: string) {
+    if (output.includes('onRunStart')) {
+      this.status.running('Running tests')
+    }
+    if (output.includes('onRunComplete')) {
+      this.status.stopped()
+      this.parsingTestFile = false
+      this.channel.clear()
```
connectdotz: hmmm... we probably don't want to clear the output channel at `onRunComplete` since that's when we needed to look at the test results. I think it might be better to do this at the beginning of each run,i.e. `onRunStart`, to clear the previous run's output instead.

src/JestExt.ts
```diff
@@ -450,12 +451,11 @@ export class JestExt {
 
   private shouldIgnoreOutput(text: string): boolean {
     // this fails when snapshots change - to be revised - returning always false for now
-    return text.includes('Watch Usage')
+    return text.includes('Watch Usage') || text.includes('onRunComplete') || text.includes('onRunStart')
   }
 
   private testsHaveStartedRunning() {
     this.channel.clear()
-    this.status.running('initial full test run')
```
connectdotz: you probably don't this method anymore. But it would be nice to keep the 'initial full test run' message, which I believe  can be done at  `onRunStart` by examing the jestProcess's watchMode.

connectdotz:
just published [jest-editor-support@27.1.0](https://www.npmjs.com/package/jest-editor-support), feel free to update this PR to utilize the new reporters support you added 👍 there.

src/JestExt.ts
jmarceli: I completely remove `testsHaveStartedRunning()` method as it is no longer required. In my opinion "initial full test run" message was also a bit redundant as just after the start it was replaced by "Running tests" message. Please let me know how do you like that approach. I just prefer a simpler code over another message in the status bar.

src/JestProcessManagement/JestProcess.ts
```diff
@@ -67,8 +71,13 @@ export class JestProcess {
     this.stopResolveCallback = null
     let exited = false
 
+    const extensionPath = vscode.extensions.getExtension(extensionId).extensionPath
+    const reporterPath = join(extensionPath, 'out', 'reporter.js')
+
     const options = {
       noColor: true,
+      shell: platform() === 'win32',
```
connectdotz: `shell` is deprecated

connectdotz:
btw, did you also try this on windows? Since there are some path related changes, worth to be extra cautious...

connectdotz:
@jmarceli I am planning 3.1.0 release next week, would like to include this PR if you have a few minutes to resolve the conflict and address the remaining minor issues. Not sure if you have windows to test this, if not, @stephtr do you think you can do a quick test to ensure this worked in windows as well?

jmarceli:
Hi, thanks for letting me know I will look at it during the weekend.

src/JestProcessManagement/JestProcess.ts
jmarceli: Removed

CHANGELOG.md
```diff
@@ -18,308 +18,311 @@ Bug-fixes within the same version aren't needed
 * quoting default "pathToJest" to preserve special characters, if any. - @connectdotz
 * Fix for when `branch.end.column` is `null`. [@garyking](https://github.com/garyking)
 * Fix several jest coverage issues and improve usability. - @connectdotz
+* Add "Running tests" message to the StatusBar - [@jmarceli](https://github.com/jmarceli)
 
 -->
 
 ### 3.0.1
 
-* use webpack for compilation, resulting in a much smaller extension size - stephtr
```
stephtr: Was that change (change stars to dashes) made deliberately?

jmarceli: I'm sorry that was prettier autoformat in my VS Code, I will revert it in a minute.

stephtr:
Thanks jmarceli, that's a nice change!
I have just tested it on Windows and so far I didn't notice any issue.

CHANGELOG.md
jmarceli: Done

henryjw:
@jmarceli Thanks for implementing this!

Issue #489, opened 2019-07-30
Jest not working because of spaces in project path


pardhu225:
### Environment

1. `node -v`: v8.15.1
2. `npm -v`: 5.6.0
3. `npm ls jest`:
```
C:\Users\Pardhu Chowdary
--` (empty)
```
4. your vscode-jest settings if customized:
 `"jest.pathToJest": "npm run test --"`

5. Operating system: Windows 10

### Prerequisite
- are you able to run jest test from command line?
No. When I run just `jest` it gives `bash: jest: command not found`. I am using git bash btw.
- how do you run your tests from command line? 
`npm run test`
Although I'm not able to run `jest` from command line, `npm run test` works fine (detects and runs all the tests)

### Steps to Reproduce

1. Installed Jest extension via marketplace.
2. Added `"jest.pathToJest": "npm run test --"` to settings.json.
3. Output in the output panel (ctrl+shift+u) when it runs:
>&gt; expense-manager@1.0.0 test e:\Dev\expense-manager
>&gt; jest --coverage --verbose "--testLocationInResults" "--json" "--useStderr" "--outputFile" "C:\Users\Pardhu" "Chowdary\AppData\Local\Temp\jest_runner_expense_manager.json" "--no-color"
>
>
>No tests found, exiting with code 1
>Run with `--passWithNoTests` to exit with code 0
>In E:\Dev\expense-manager
>  46 files checked.
>  testMatch:  - 0 matches
>  testPathIgnorePatterns: \\node_modules\\ - 46 matches
>  testRegex: (\\\_\_tests\_\_\\.*|(\.|\\)(test|spec))\.ts$ - 1 match
>Pattern: Chowdary\AppData\Local\Temp\jest_runner_expense_manager.json - 0 matches

I'm pretty sure this is caused due to the space in the path: 
`C:\Users\Pardhu Chowdary\AppData\Local\Temp\jest_runner_expense_manager.json`
Tried the fix as suggested in #342 
### Relevant Debug Info

Partial package.json:
```
  "main": "dist/src",
  "scripts": {
    "prebuild": "tslint -c tslint.json -p tsconfig.json --fix",
    "build": "tsc",
    "prestart": "npm run build",
    "start": "node .",
    "test": "jest --coverage --verbose",
    "dev": "tsc-watch --onSuccess \"node ./dist/index.js\""
  },
  "dependencies": {
    "body-parser": "^1.19.0",
    "colors": "^1.3.3",
[...]

Issue #426, opened 2019-02-22
Plugin won't run on Windows when project is in a path with a space in it.


paulfelton:
### Environment

1. `node -v`: [v8.11.3]
1. `npm -v`: [v6.8.0]
1. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): [n/a]

1. Operating system: [Windows 8.1]

### Prerequisite
- are you able to run jest test from command line? [yes]
- how do you run your tests from command line? (for example: `npm run test` or `node_modules/.bin/jest`) [npm run test]

### Steps to Reproduce

1. Place your project in a folder containing a space, eg: e:\My Websites\myproject.
Plugin attempts to run multiple times before giving up, saying it can't find 'e:\My'.

### Expected Behavior

Plugin should be able to handle Windows paths

### Actual Behavior

Complains it can't find 'e:\My'

--- 

The fastest (and the most fun) way to resolve the issue is to submit a pull-request yourself. If you are interested, feel free to check out the [contribution guide](https://github.com/jest-community/vscode-jest/CONTRIBUTING.md), we look forward to seeing your PR...

beatgates:
Same issue as #360, #342
Work-around was to add in vs code settings "jest.pathToJest": "npm run test --"
I don't believe #342 should be closed with just a work-around?

Issue #360, opened 2018-08-13
vscode-jest stopping at launch of VSCode 


talentedjerkkk23:
### Environment

1. `node -v`: v9.4.0 
1. `npm -v`: 6.3.0
1. VSCode: 1.25.1
1. Operating system: Windows 7 x64

### Prerequisite
- are you able to run jest test from command line? yes
- how do yo run your tests from command line? yarn test

###  Steps to Reproduce
1. Install Jest as dev-dependency
2. Create jest.config.js with following:
`module.exports = {
    verbose: true,
    testURL: "http://localhost/",
} `

### Expected Behavior
vscode-jest should work 

### Actual Behavior
After opening project with jest in VSCode:
 - vscode-jest stopped and printed the following: `Starting Jest in Watch mode failed too many times and has been stopped.`
- after reloading VSCode 100+ times - vscode-jest started working properly
- after another VSCode window reload vscode-jest stopped again

Flavien-Pensato:
Did you try without `testURL` ?

talentedjerkkk23:
@Flavien-Pensato yes, same thing

connectdotz:
a few things to try:
1. do [self-diagnosis](https://github.com/jest-community/vscode-jest#plugin-not-running-as-expect-try-self-diagnosis): turning on the debug mode and see what command vscode-jest used to spaw your jest
1. if vscode-jest is not using `yarn test`, you can override it by setting the `jest.pathToJest`.

talentedjerkkk23:
@connectdotz extension using `yarn jest` as i can see using debug mode.

oleersoy:
I'm having the same problem.  

Jest is trying to scan the `node_modules` directory for tests even though I tell it to ignore it in `package.json`:

```
  "jest": {
    "testPathIgnorePatterns": [".test", "node_modules"]
  },

```
This is the console output:

```
jest-haste-map: @providesModule naming collision:
  Duplicate module name: prototype project
  Paths: /home/ole/SuperflyCSS/cli/test/sandbox/d5245d4c-b921-4e2a-9962-c82238ce0419/test/newproject/package.json collides with /home/ole/SuperflyCSS/cli/test/sandbox/7ab8e351-79e5-49c5-8f3c-ce0dd44d87c4/test/newproject/package.json

This warning is caused by a @providesModule declaration with the same name across two different files.

internal/fs/watchers.js:173
    throw error;
    ^

[...]

PR #522, opened 2019-11-21
Quote default path to jest

connectdotz:
this is to follow up jest-community/jest-editor-support#21:

with `jest-editor-support` 27.0.0 now able to pass raw pathToJest to the spawn call, users can embed escape characters or quotes when they customize `jest.pathToJest` and our generated default jestToPath should do the same in case there is escape sequence in user's path. 

related issues:
#342
#360
#426
#489


connectdotz:
replaced with #523

PR #523, opened 2019-11-21
quoting default pathToJest

connectdotz:
this is to follow up jest-community/jest-editor-support#21:

with `jest-editor-support` 27.0.0 now able to pass raw pathToJest to the spawn call, users can embed escape characters or quotes when they customize `jest.pathToJest` and our generated default jestToPath should do the same in case there is escape sequence in user's path. 

related issues:
#342
#360
#426
#489

### DIFF
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -79,7 +79,8 @@ export function pathToJest({ pathToJest, rootPath }: IPluginResourceSettings) {
     return 'npm test --'
   }
 
-  return getLocalPathForExecutable(rootPath, 'jest') || 'jest' + nodeBinExtension
+  const p = getLocalPathForExecutable(rootPath, 'jest') || 'jest' + nodeBinExtension
+  return `"${p}"`
 }
 
 /**
--- a/src/helpers.ts
+++ b/src/helpers.ts
@@ -96,34 +96,6 @@ export function pathToConfig(pluginSettings: IPluginResourceSettings) {
   return ''
 }
 
-export function pathToJestPackageJSON(pluginSettings: IPluginResourceSettings): string | null {
-  let pathToNodeModules = join(pluginSettings.rootPath, 'node_modules')
-
-  if (pluginSettings.pathToJest) {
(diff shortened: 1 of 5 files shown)


connectdotz:
@omjadas @stephtr @seanpoulter does any of you have a few minutes to review so we can close this issue and cut a new release soon...

src/helpers.ts
```diff
@@ -79,7 +79,8 @@ export function pathToJest({ pathToJest, rootPath }: IPluginResourceSettings) {
     return 'npm test --'
   }
 
-  return getLocalPathForExecutable(rootPath, 'jest') || 'jest' + nodeBinExtension
+  const p = getLocalPathForExecutable(rootPath, 'jest') || 'jest' + nodeBinExtension
```
stephtr: I don't have it in mind, if we always use the shell, do we need `nodeBinExtension` on Windows?

stephtr:
Beside the shell option it looks good to me.

src/helpers.ts
connectdotz: I don't have a windows machine to verify, if it worked without, I can remove it... does it matter which shell is used? I could imagine users might want to pass their favor shell in the future...

connectdotz:
@omjadas thanks for reminding me about the deprecated shell option, it's now fixed.

@stephtr I am glad you are back! I agree we should remove the `nodeBinExtension` for windows if we could, but would definitely need to test it before committing. Since I didn't have a way to verify this, I think it's best to leave it as it is until maybe you or someone else to submit a separate PR for windows.

connectdotz:
@stephtr @omjadas is there any other change that should be considered? If not please approve otherwise Github blocks merging.

stephtr:
> I agree we should remove the nodeBinExtension for windows if we could, but would definitely need to test it before committing. Since I didn't have a way to verify this, I think it's best to leave it as it is until maybe you or someone else to submit a separate PR for windows.

That's true, especially since we already had a few issues related to that.

Issue #499, opened 2019-09-17
Coverage not shown on typescript


jeantil:
### Environment

1. `node -v`: v10.16.0
1. `npm -v`: 6.9.0  /`yarn -v` 1.17.3
1. `npm ls jest` 
   ```
    npm ls jest
   180@1.0.0 /Users/jean/dev/startups/yupwego/src/vscode-jest-499
   └── jest@24.9.0
   ```
1. `yarn why jest`
   ```
   yarn why jest
   yarn why v1.17.3
   [1/4] 🤔  Why do we have the module "jest"...?
   [2/4] 🚚  Initialising dependency graph...
   [3/4] 🔍  Finding dependency...
   [4/4] 🚡  Calculating file sizes...
   => Found "jest@24.9.0"
   info Has been hoisted to "jest"
   info This module exists because it's specified in "devDependencies".
   info Disk size without dependencies: "48KB"
   info Disk size with unique dependencies: "276KB"
   info Disk size with transitive dependencies: "51.21MB"
   info Number of shared dependencies: 43
   ✨  Done in 0.57s.
   ``` 
1. your vscode-jest settings if customized:
    - "jest.showCoverageOnLoad": true

1. Operating system: OSX Mojave 10.14.6
1. installed vscode extensions
  ```
  code --list-extensions
  AndrewRazumovsky.vscode-styled-jsx-languageserver
  andys8.jest-snippets
  Arjun.swagger-viewer
  blanu.vscode-styled-jsx
  christian-kohler.npm-intellisense
  dbaeumer.vscode-eslint
  DotJoshJohnson.xml
  eamodio.gitlens
  EditorConfig.EditorConfig
  eg2.vscode-npm-script
  eriklynd.json-tools
  esbenp.prettier-vscode
  file-icons.file-icons
  joaompinto.asciidoctor-vscode
  lightbend.vscode-sbt-scala
  mrorz.language-gettext
  ms-azuretools.vscode-docker
  Orta.vscode-jest
  PKief.material-icon-theme
  ryu1kn.partial-diff
  scala-lang.scala
  scalameta.metals
  skyran.js-jsx-snippets
  stkb.rewrap
  vscode-icons-team.vscode-icons
  wayou.vscode-todo-highlight
  xabikos.ReactSnippets
  ```
### Prerequisite
- are you able to run jest test from command line? 
  ```
  yarn jest
  yarn run v1.17.3
[...]

Issue #490, opened 2019-08-12
Unstable vscode-jest coverage overlay


tadaslape-vs:
### Environment

1. `node -v`: v12.4.0
1. `npm -v`: 6.9.0
1. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): react-scripts@3.0.1
1. your vscode-jest settings if customized:
    - jest.pathToJest? `npm run vscode-jest --` (react-scripts test --coverage --watchAll=true)
    - jest.pathToConfig? -
    - anything else that you think might be relevant? -

1. Operating system: Windows 10 Enterprise

### Prerequisite
- are you able to run jest test from command line? Yes.
- how do you run your tests from command line? (for example: `npm run test` or `node_modules/.bin/jest`) `npm run jest`

### Steps to Reproduce

1. Edit command linked in `jest.pathToJest` to `react-scripts test --coverage --watchAll=false`.
2. `Jest: Toggle Coverage Overlay` works fine when enabling/disabling.

3. Edit command linked in `jest.pathToJest` to `react-scripts test --coverage --watchAll=true`.
4. `Jest: Toggle Coverage Overlay` disables overlay once and enabling it does not work anymore until reloading VS Code window. Also, changing between files (possibly due to new errors being detected) makes the overlay also stop working.

### Expected Behavior

The coverage overlay should be consistent and enable/disable just fine without hiding itself on it's own.

### Actual Behavior

The coverage overlay is very unstable. It disappears on itself or disabling it blocks from enabling it without reloading VS Code. The inconsistency comes with missing flag ` --watchAll=false` in path to Jest command.

Issue #488, opened 2019-07-28
Coverage overlay disappears after switching tabs


garyking:
### Environment

1. `node -v`: v10.16.0
1. `npm -v`: 6.10.2
1. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): jest@24.8.0
1. your vscode-jest settings if customized:
    - jest.pathToJest? "node_modules/.bin/jest"

1. Operating system: MacOS 10.14.5

### Prerequisite
- are you able to run jest test from command line? Yes
- how do you run your tests from command line? (for example: `npm run test` or `node_modules/.bin/jest`) jest

### Behavior

Whenever I want the coverage overlay to appear, I have to stop the Jest runner, then start it again. It should not be necessary to do this, but for now, it would be nice to have a Restart option instead.

So when I start Jest, then the coverage appears for the current file. The summary at the top of the file always appears for a split second. If I switch tabs to another file, then the coverage overlay disappears.

I've attached a GIF of the issue.

![Jest VS Code - Coverage overlay](https://user-images.githubusercontent.com/343940/62013292-4984e880-b15e-11e9-9619-356edc7d2071.gif)

garyking:
Fixed per [this comment of mine](https://github.com/jest-community/vscode-jest/issues/370#issuecomment-554595209).

Issue #449, opened 2019-04-22
Question: Sidebar shows errors before runnning


donifer:
### Environment

1. `node -v`: 10.15.1
1. `npm -v`: 6.4.1

1. Operating system: macOS

### Prerequisite
- are you able to run jest test from command line? ues
- how do you run your tests from command line? `yarn run test`

### Question
When I first open a repo in vscode all my test files are red in the sidenav until I manually open each one of them. Is this expected? Is there a workaround? Not sure what I'm doing wrong.

PR #527, opened 2019-11-27
Better coverage

connectdotz:
fix a few test coverage issues (#429, #449, #488, #490, #499) and improve usability:

- the test coverage map only contains the tests in the latest run, which means in watch mode we lost the map generated during the earlier/initial test run. This leads to people had to pass `--watchAll` or restart the plugin in order to see coverage.
  - use `merge` to accumulate the maps instead of overriding.

- current coverage features display info from coverage-map but they do not help to generate the map. Users still need to pass the `--coverage` flag manually in either `jest.pathToJest`, jest config or package.json script to make it work. It is counter-intuitive and inefficient (why generate the map when toggle off the display, or toggle on display but do not generate the map?).
  - Giving `jest-editor-support` already has a `collectCoverage` option, we can easily wire it up with the existing features to provide the single-command, true-dynamic coverage experience out-of-the-box. 

- there is no indicator for coverage state, which can be confusing when the file is 100% covered. 
  - Added new "mode" icons in the status bar, currently shows "watch" and "coverage" state.

Screen Shot 2019-11-27 at 11 07 34 PM


connectdotz:
@seanpoulter @stephtr @garyking when you guys get a chance, please take a look... hopefully, this will make our coverage support more intuitive going forward...

connectdotz:
ping... anybody?

src/StatusBar.ts
```diff
@@ -240,6 +246,22 @@ export class StatusBar {
...
+      switch (m) {
+        case 'coverage':
+          return '$(color-mode)'
+        case 'watch':
+          return '$(eye)'
+        default:
+          throw new Error(`unrecognized mode: ${m}`)
+      }
+    })
+    return modesStrings.length > 0 ? `${modesStrings.join(' ')}` : ''
```
stephtr: Why not simply `return modesStrings.join(' ')`?

src/StatusBar.ts
```diff
@@ -92,20 +95,20 @@ export class StatusBar {
...
       },
-      failed: (details?: string) => {
-        this.request(source, 'failed', details)
+      failed: (details?: string, modes?: Mode[]) => {
+        this.request(source, 'failed', details, modes)
       },
-      stopped: (details?: string) => {
-        this.request(source, 'stopped', details)
+      stopped: (details?: string, modes?: Mode[]) => {
+        this.request(source, 'stopped', details, modes)
```
stephtr: This pattern feels a bit strange. What about instead adding a method to JestExt which calls `getStatusModes` and `request`?

connectdotz: > What about instead adding a method to JestExt which calls getStatusModes and request?
right now mode and status are combined into a single string, if we make them separate calls, extra logic is needed to remember the missing info.

However, having said that, we could simplify the API by removing modes from all binding method except `initial` and injecting them in the wrapper within...

src/StatusBar.ts
connectdotz: indeed

src/StatusBar.ts
connectdotz: turns out the new API I was thinking about is even more cumbersome... maybe I should understand a bit more your concern so we can come up with a better approach...

connectdotz:
> Maybe one can extend it in a future PR with a status bar button for switching the coverageOverlay.

that is a great idea, but you are right let's address it in another PR.

src/StatusBar.ts
stephtr: My approach would have been https://github.com/connectdotz/vscode-jest/pull/1.

PR #533, opened 2019-12-14
document: Add install through Visual Studio Marketplace

tarepan:
Add alternative install method to README.md, which use Visual Studio Marketplace.  
Simple 'search' in VSCode's extention tab force people judge which extetion is this extention.  
Visual Studio Marketplace enable installation of specified extention, so remove this pain.


stephtr:
That's a good idea, in my opinion we should even favour this approach over the search procedure.

connectdotz:
thanks, @tarepan!

PR #536, opened 2019-12-15
Fix StatusBarItem in case no active editor is open on launch

stephtr:
When having no active text editor opened while Jest starts, the StatusBarItem shows `Jest status of 'undefined'` as a tooltip and can't be clicked to open the summary.
Since there's no ambiguity when there is only a single WorkspaceFolder (and the summary view only takes over for more than one WorkspaceFolder), it is optimal to take that WorkspaceFolder as `activeFolder`.

### DIFF
--- a/src/StatusBar.ts
+++ b/src/StatusBar.ts
@@ -127,14 +127,14 @@ export class StatusBar {
   }
 
   private get activeFolder() {
+    if (!this._activeFolder && vscode.workspace.workspaceFolders.length === 1) {
+      this._activeFolder = vscode.workspace.workspaceFolders[0].name
+    }
     if (!this._activeFolder && vscode.window.activeTextEditor) {
       const folder = vscode.workspace.getWorkspaceFolder(vscode.window.activeTextEditor.document.uri)
       if (folder) {
         this._activeFolder = folder.name
       }
-    }
-    if (!this._activeFolder && vscode.workspace.workspaceFolders.length === 1) {
-      this._activeFolder = vscode.workspace.workspaceFolders[0].name
     }
     return this._activeFolder
   }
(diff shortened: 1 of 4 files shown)


src/StatusBar.ts
```diff
@@ -127,6 +127,9 @@ export class StatusBar {
   }
 
   private get activeFolder() {
+    if (!this._activeFolder && vscode.workspace.workspaceFolders.length === 1) {
+      this._activeFolder = vscode.workspace.workspaceFolders[0].name
+    }
     if (!this._activeFolder && vscode.window.activeTextEditor) {
       const folder = vscode.workspace.getWorkspaceFolder(vscode.window.activeTextEditor.document.uri)
       if (folder) {
```
connectdotz: maybe combine these 2 conditions under `if (!this._activeFolder) {...}` to reduce a bit mental overhead? Also, did you mean to have an `else` after the first condition?

stephtr: Either there have to be two `if(!this._activeFolder)` or an `else`. I'll change it.

stephtr:
> just some minor points below. The test coverage has decreased thus the ci failure, you might want to add some test to test the new condition.

I wanted to do that but unfortunately that's a private method and an implementation detail. I'll try to find a more meaningful test today. Maybe you can wait with the 3.1 release until then.

connectdotz:
no problem, I will wait.

tests/JestProcessManagement/JestProcess.test.ts
```diff
@@ -35,7 +36,7 @@ describe('JestProcess', () => {
       new JestProcess(projectWorkspaceMock)
       expect(runnerMock).toHaveBeenCalledWith(undefined, {
         noColor: true,
-        reporters: ['default', '/my/vscode/extensions/out/reporter.js'],
+        reporters: ['default', normalize('/my/vscode/extensions/out/reporter.js')],
```
stephtr: This change fixes the test failing on Windows (where the slashes got only partially changed to backslashes).

PR #547, opened 2020-01-14
quote reporter path

### LINKED ISSUE #545: After installing the extension, it is not able to start
### Environment

1. `node -v`: v12.13.0
1. `npm -v`: 6.12.0
1. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): jest@24.9.0

1. Operating system: Windows 10

### Prerequisite
- are you able to run jest test from command line? - Yes, I am.
- how do you run your tests from command line? (for example: `npm run test` or `node_modules/.bin/jest`) npm run test

### Steps to Reproduce
I simply installed your extension after creating a basic react project and received the following error:
image

My username is 'Tony Troeff' but I see that the path of the module name is only "C:\Users\Tony" and this is the reason why such a reporter is not found, I suppose.

### Expected Behavior

After installing your extension, it should be able to start.

connectdotz:
reporter path needs to be in quotes to make sure whitespace in user path won't truncate the command. 

fix #545

### DIFF
--- a/src/JestProcessManagement/JestProcess.ts
+++ b/src/JestProcessManagement/JestProcess.ts
@@ -75,7 +75,7 @@ export class JestProcess {
 
     const options = {
       noColor: true,
-      reporters: ['default', reporterPath],
+      reporters: ['default', `'${reporterPath}'`],
     }
     this.runner = new Runner(this.projectWorkspace, options)
 
--- a/src/JestProcessManagement/JestProcess.ts
+++ b/src/JestProcessManagement/JestProcess.ts
@@ -75,7 +75,7 @@ export class JestProcess {
 
     const options = {
       noColor: true,
-      reporters: ['default', `'${reporterPath}'`],
+      reporters: ['default', `"${reporterPath}"`],
     }
     this.runner = new Runner(this.projectWorkspace, options)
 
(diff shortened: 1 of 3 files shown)


connectdotz:
This is an urgent bug fix as there is no workaround, @stephtr if you can do a quick review maybe we can get it out today...

stephtr:
I guess we don't care about quotemarks in the `extensionPath`? On Windows I would prefer doublequotes since only they are forbidden in file names.

connectdotz:
Good point 👍

src/JestProcessManagement/JestProcess.ts
```diff
@@ -75,7 +75,7 @@ export class JestProcess {
 
     const options = {
       noColor: true,
-      reporters: ['default', reporterPath],
+      reporters: ['default', `'${reporterPath}'`],
```
stephtr: Or should we do a `${reporterPath.replace("'","\\'")}`? If not, I'd be also fine with the current solution.

stephtr:
Wouldn't it be better to quote it where the `reporters` property gets consumed?

connectdotz:
Do you mean in `jest-editor-support` or maybe `jest`? I think the current thinking is to leave it to the users to pass in whatever they see fit instead of hard coding on the consuming  end

stephtr:
> Do you mean in `jest-editor-support` or maybe `jest`? I think the current thinking is to leave it to the users to pass in whatever they see fit instead of hard coding on the consuming end

Ok. I was just slightly bemused by an array having to be quoted, since I then would prefer passing an already joined string instead. But the decision has been already made anyway.

PR #566, opened 2020-04-01
Add the ability to detect CRA apps even with cross-env in use

stephtr:
As [noted in #244](https://github.com/jest-community/vscode-jest/issues/244#issuecomment-456709258) when using specific generators different than `create-react-app`, `react-scripts` gets prepended by `"cross-env CI=true"`, which fails in our `isCreateReactAppTestCommand` function.
This PR takes care of `cross-env` and fixes the last few comments in #244.


stephtr:
The CI failure is unrelated to the change.

connectdotz:
I am worrying that there will be many more possible variations that we will end up adding more and more logic to handle them... but where do we draw the line? how about only supporting the "default" CRA test script? If users end up customizing their test script, it would made sense to customize the `jest.pathToConfig` as well. Most importantly, we should document what we do support to set the proper expectation going forward.

BTW, the build failure is due to missing CHANGELOG update...

stephtr:
It's a good question where to draw the line. If there is such a prominent app like Visual Studio, which uses `cross-env` in its template, in addition to `cross-env` being quite popular, for me adding six lines of code would be kind of alright, even though I'm also unsure.

> BTW, the build failure is due to missing CHANGELOG update...

Thanks, that's already at least the second time that I forgot that 🤦‍♂️

connectdotz:
@stephtr I see where you come from... let's take a step back, what is all this about? we want to know if we should call the test script instead of jest binary... Actually when you come to think about it, if there is a "test" script in the package.json, wouldn't it very possible that we should just use it regardless if it is CRA? Of course, it is probably not 100%, thus the hesitation...

IMHO, the root cause of all these struggles is that we try to decide all these behind the scene without any user input. What if we give users a chance to participate...? I remember we had a similar conversation before.

Say, we have a command, setting-helper, which examines `jest.pathToJest`, jest bin, and if there is a test script present in package.json (we already all these), we then gave users choices, maybe in priority: jest.pathToJest > test script > jest binary to use; or to provide their own manually. We can then open the workspace settings and update `jest.pathToJest` with their choice.

On every startup, if `jest.pathToJest` is set, we will just use it, otherwise, run this setting-helper again. (this command can also be activated from the command palette so users can change it anytime on-demand as well).

We can use the system message mechanism that we used for reporting too-many-error today, or perhaps there are better vscode UI choices...

The confusion of the startup command is probably the biggest sore spot for this extension, by doing this we not only removed the "wrong guess" but also introduces the customization early on. I think it will make this extension so much more awesome! 

What do you think? want to implement it in this PR? we could start a `4.0` to rollout a few major improvements, this will be on the top of that list.

stephtr:
> Actually when you come to think about it, if there is a "test" script in the package.json, wouldn't it very possible that we should just use it regardless if it is CRA? Of course, it is probably not 100%, thus the hesitation...

For just running the tests that could be an option. One would still have to think about for example Yarn workspaces. However, for debugging tests, at least as far as I have it in mind, CRA and pure jest setups need different handling, even if one is just calling `npm test`.

> Say, we have a command, setting-helper, which examines jest.pathToJest, jest bin, and if there is a test script present in package.json (we already have all these logics), we then gave users choices, maybe in priority: jest.pathToJest > test script > jest binary to use; or to provide their own manually. We can then open the workspace settings and update jest.pathToJest with their choice.

That would be an interesting option! I would be thinking about something like the C# extension:
image

> What do you think? want to implement it in this PR? we could start a 4.0 to rollout a few major improvements, this will be on the top of that list.

After we think about it in detail (also how the debug configuration fits into such a user experience, that could be a great feature! Unfortunately, since a few months I'm struggling with time and that won't change for the next few months. Starting such a PR simply doesn't fit my current time constraints.

In the meantime I simplified CRA detection to just look whether the test string contains (instead of starts with) for example `react-scripts test`. That realistically shouldn't give us false positives. In my opinion we could also get rid of the [folder lookup fallback](https://github.com/jest-community/vscode-jest/blob/1d522a378cc1cdb3e9a11743f5a4558f0ab0b5fc/src/helpers.ts#L60). All of that would be just temporarily, until we have the code changes for 4.0.

src/helpers.ts
```diff
@@ -48,7 +48,8 @@ export function getTestCommand(rootPath: string): string | undefined | null {
  * Checks if the supplied test command could have been generated by create-react-app
  */
 export function isCreateReactAppTestCommand(testCommand: string): boolean {
-  return testCommand && createReactAppBinaryNames.some(binary => testCommand.indexOf(binary + ' test') === 0)
+  testCommand = ` ${testCommand} `
+  return createReactAppBinaryNames.some(binary => testCommand.includes(` ${binary} test `))
```
connectdotz, line 52: hmm... would it be better to use word-boundary instead of just space? what if the command has quotes, tabs, or semicolon around the test token instead?

a minor performance consideration: if testCommand is falsy (undefined, empty etc), it would be more efficient to short circuit it out of the matching...

connectdotz:
> That would be an interesting option! I'm imagining something like the C# extension:

yeah, something along that line...

> Unfortunately, since a few months I'm struggling with time and that won't change for the next few months. Starting such a PR simply doesn't fit my current time constraints.

I know that feeling, I was on the same boat last month. That's fine, do what you can and hope you get out of the hole soon 😏

src/helpers.ts
stephtr: Or maybe it would be easier to not even check what precedes and succeeds the `react-scripts test` command. `fake-react-scripts tester` would then be a false positive, but that's manageable. We are only doing heuristics and users can customize their settings.

connectdotz: yeah, I think this is a reasonable improvement for our current situation...

PR #567, opened 2020-04-02
Use .bin jest

glentakahashi:
If we use node_modules/.bin/jest this plugin will work in yarn workspaces with a hoisted jest without requiring modifications


connectdotz:
will this work for npm and other non-yarn environments?

stephtr:
That won't work so easily. `jest/bin/jest` is a JS file, while `.bin/jest` is a shell file, on Windows it therefore doesn't attach the debugger. However, apparently the [VS Code recipe](https://github.com/microsoft/vscode-recipes/tree/master/debugging-jest-tests) for jest has changed. We could use that for supporting yarn workspaces at least on linux.

connectdotz:
@stephtr I see, so we can add a separate program for windows like the vscode recipe suggested. 

@glentakahashi do you know if this change will work for npm? Does npm also create the hoisted link in the .bin directory? Judging from the receipt, it seems to imply that, but just want to confirm...

stephtr:
> I see, so we can add a separate program for windows like the vscode recipe suggested.

I would suggest updating our default debug configuration. `package.json` is just one part, the other part resides in [DebugConfigurationProvider.ts](https://github.com/jest-community/vscode-jest/blob/e3a3056b38da16ee73e6e02203516eb90cee8865/src/DebugConfigurationProvider.ts#L48).

> Does npm also create the hoisted link in the .bin directory?

Yes, that's part of the [specs](https://docs.npmjs.com/files/folders#executables).

glentakahashi:
Ah okay, so would the new proposal be to use .bin/jest in the package.json + DebugConfigurationProvider.ts for unix, and for both to also add a specific windows component to still use the jest/bin/jest?

connectdotz:
that sounded right...

seanpoulter:
I'm going to go way back to the beginning. What problem is this solving? What is "a hoisted jest" and when would we end up with one?

glentakahashi:
When using yarn workspaces (https://classic.yarnpkg.com/en/docs/workspaces/) dependencies from different workspaces will be hoisted to the top level directory. e.g.:

`root/packages/a/package.json` -> `jest: ^26.0`
`root/packages/b/package.json` -> `jest: ^26.1`

will result in
a file structure of
`root/node_modules/jest/`{all the normal jest files}
`root/packages/a/node_modules/.bin/jest` -> symlink to `../../../../node_modules/.bin/jest`
`root/packages/b/node_modules/.bin/jest` -> symlink to `../../../../node_modules/.bin/jest`

So when opening up vscode in a package directory such as `code root/packages/a`, vscode-jest will not work since it currently references `node_modules/jest/bin/jest`

seanpoulter:
Ah ha, interesting! Thanks Glen, I really appreciate the thoughtful explanation. I've used npm, yarn, and lerna and haven't run into workspaces or the missing `packages/__/node_modules/jest/bin/jest` yet.

seanpoulter:
Alright, I'm on board that this a problem that we need to solve. Is there a way to check if the package is hoisted and use that JavaScript file with our existing "debugging recipe" and code?

I'm thinking about an asynchronous function that will find the module path:
- In the current package, read where the symbolic link `node_modules/.bin/jest` resolves to
- We'll need (at least) three branches:
    - If the file is found:
        - If the file is a symbolic link, resolve the path to jest/main.js from `.bin/jest`
        - If the file is not a symbolic link, keep the status quo
    - If the file is not found, status quo (we can fail downstream)

Does that simplify and of the OS-specific, case-specific mess we're in?

connectdotz:
I think the idea is by adopting `.bin/jest` we don't need to worry about hoisting variation, the pm will take care of that for us. And by adding the `windows` option in the launch.json, along with the non-widnows (per  [VS Code recipe](https://github.com/microsoft/vscode-recipes/tree/master/debugging-jest-tests#configure-launchjson-file-for-your-test-framework)), we also don't need to do any runtime checking to figure out the platform/OS etc. 

I think the fact that all these can be accomplished in a `static` receipt change, i.e. no need to do any runtime check for hoist/path/os, is pretty sweet! 😄

seanpoulter:
Ah, well I'm happy I misunderstood the conversation so far then. :D 

When I read the following, it sounded like we were going to be spreading out the changes in more parts of the system than necessary (emphasis my own):

> DebugConfigurationProvider.ts **for unix**, and for both to also add a **specific windows** component to still use the jest/bin/jest

I'll leave it to you three capable folks! 👏

glentakahashi:
Hey all, sorry for the late reply but I've updated it per our conversation!

glentakahashi:
Okay tested on both windows + mac that the updated configurations work and debugging still works as well.

connectdotz:
@glentakahashi thanks for this PR! ci build failed is due to lack of changelog update, once it is fixed I will merge it.

glentakahashi:
@connectdotz done!

stephtr:
Thank you for the PR!
On Windows basically nothing has changed (the "issue" is still present). However, when I looked into it I haven't found a better solution. Since this PR resembles [the configuration suggested by jest's documentation](https://jestjs.io/docs/en/troubleshooting.html#debugging-in-vs-code), I would nevertheless keep it as is.
As a side note: When working with Yarn workspaces I would anyway recommend just opening the workspace's root.

PR #577, opened 2020-04-29
Use --watchAll when using submodules

### LINKED ISSUE #558: Option to run always with `--watchAll` instead of `--watch`
### Environment

1. `node -v`: v12.13.1
1. `npm -v`: 6.13.6
1. Operating system: macOS Mojave

### Prerequisite
- are you able to run jest test from command line? Only using `--watchAll`
- how do you run your tests from command line? (for example: `npm run test` or `node_modules/.bin/jest`) `pnpx jest --watchAll`

### Steps to Reproduce

In a monorepo using git submodules and a multi-project setup (`jest.config.js` in the root using the `projects` option to run the test suites of all packages), running Jest in watch mode (`jest --watch`) always errors out with something along the lines of:

```
Determining test suites to run...

  ● Test suite failed to run

    fatal: ../packages/a-dependency-outside-the-submodule: '../packages/a-dependency-outside-the-submodule' is outside repository
```

This is a variant of [a known Jest issue](https://github.com/facebook/jest/issues/4419), and the usual workaround is to run Jest with `--watchAll` instead of `--watch`.

vscode-jest aways errors out, though, because it tries to use `--watch` and there is no way (that I could find) for force it to run Jest with `--watchAll`.

### Expected Behavior

I would like to have an option to run Jest with vscode-jest using `watchAll` instead of `watch`, or having the extension detect this kind of error and behave accordingly.

lordofthelake:
Updates the RegExp used to detect when jest should be restarted using `--watchAll`.
Closes #558.

### DIFF
--- a/src/Jest/index.ts
+++ b/src/Jest/index.ts
@@ -5,7 +5,7 @@ export enum WatchMode {
 }
 
 const IS_OUTSIDE_REPOSITORY_REGEXP = /Test suite failed to run[\s\S]*fatal:[\s\S]*is outside repository/im;
-const WATCH_IS_NOT_SUPPORTED_REGEXP = new RegExp('^s*--watch is not supported without git/hg, please use --watchAlls*', 'im');
+const WATCH_IS_NOT_SUPPORTED_REGEXP = /^s*--watch is not supported without git\/hg, please use --watchAlls*/im;
 
 export const isWatchNotSupported = (str = '') =>
   IS_OUTSIDE_REPOSITORY_REGEXP.test(str) || WATCH_IS_NOT_SUPPORTED_REGEXP.test(str)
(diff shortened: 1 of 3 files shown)


src/Jest/index.ts
```diff
@@ -4,5 +4,8 @@ export enum WatchMode {
   WatchAll = 'watchAll',
 }
 
+const IS_OUTSIDE_REPOSITORY_REGEXP = /Test suite failed to run[\s\S]*fatal:[\s\S]*is outside repository/im;
+const WATCH_IS_NOT_SUPPORTED_REGEXP = new RegExp('^s*--watch is not supported without git/hg, please use --watchAlls*', 'im');
```
connectdotz, line 8: I am fine either one but let's use consistent styles for these patterns

lordofthelake:
@connectdotz Thanks for the pointer for the missing changelog entry! So ✅now 😍

PR #579, opened 2020-04-29
update most dependencies to the latest

connectdotz:
partly to address the security warning, partly to address the maintainability going forward...

- upgrade all dependencies to the latest, except istanbul-lib-xxx, which requires more code change and will be handled in a separate coverage PR
- made some minor code change due to lint complaint
- migrate `vsode` dependency to `vscode-test` ([here](https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode))
- clean up some unused dependency and typing files.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "vscode-jest",
   "displayName": "Jest",
   "description": "Use Facebook's Jest With Pleasure.",
-  "version": "3.1.1",
+  "version": "3.2.0",
   "publisher": "Orta",
   "engines": {
     "vscode": "^1.23.0"
--- a/package.json
+++ b/package.json
@@ -250,7 +250,7 @@
     "jest-editor-support": "^27.2.0",
     "jest-snapshot": "^25.5.0"
   },
-  "optionalDependencies": {
+  "devDependencies": {
     "@types/jest": "^25.2.1",
     "@types/node": "^8.0.31",
     "@types/vscode": "^1.23.0",
--- a/package.json
+++ b/package.json
@@ -245,15 +245,15 @@
     "prettier-project": "yarn prettier-write -- '?(__mocks__|src|tests)/**/*.ts'"
   },
   "dependencies": {
-    "@types/jest": "^25.2.1",
-    "@types/node": "^8.0.31",
-    "@types/vscode": "^1.23.0",
     "istanbul-lib-coverage": "^1.1.1",
     "istanbul-lib-source-maps": "^1.1.0",
     "jest-editor-support": "^27.2.0",
     "jest-snapshot": "^25.5.0"
   },
   "optionalDependencies": {
+    "@types/jest": "^25.2.1",
+    "@types/node": "^8.0.31",
+    "@types/vscode": "^1.23.0",
     "coveralls": "^3.1.0",
     "danger": "^10.1.1",
     "jest": "^25.5.0",
--- a/package.json
+++ b/package.json
@@ -290,17 +290,18 @@
     "prettier-write": "yarn prettier --parser typescript --no-semi --single-quote --trailing-comma es5 --write --print-width 120",
     "prettier-project": "yarn prettier-write -- '?(__mocks__|src|tests)/**/*.ts'"
   },
-  "devDependencies": {
+  "dependencies": {
     "@types/jest": "^25.2.1",
     "@types/node": "^8.0.31",
     "@types/vscode": "^1.23.0",
-    "coveralls": "^3.1.0",
-    "danger": "^10.1.1",
     "istanbul-lib-coverage": "^1.1.1",
     "istanbul-lib-source-maps": "^1.1.0",
-    "jest": "^25.5.0",
     "jest-editor-support": "^27.2.0",
-    "jest-snapshot": "^25.5.0",
+    "jest-snapshot": "^25.5.0"
+  },
+  "devDependencies": {
+    "coveralls": "^3.1.0",
+    "jest": "^25.5.0",
     "lint-staged": "^10.2.0",
     "prettier": "^2.0.5",
     "rimraf": "^3.0.2",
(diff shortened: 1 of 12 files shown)


connectdotz:
hmmm... looks like the latest version of danger@10.1.1 uses gitlab@10.2.1 which requires node >= 10.0.0... there are multiple approaches we can take:
1. up our minimal node requirement
1. find an alternative to danger

looks like both jest and vscode still support node 8... thoughts?

connectdotz:
actually, danger is a build tool, we should be able to separate that from the regular dependencies and travis build... let me do some research and see if we can split them up, right now everything is un devDependencies, which is ... well odd.

connectdotz:
ok, I got travis to work now, but I would admit it's a bit hacky, all ears if you got better ideas...

Here is what I did:
- split the dependencies into `dependencies` and `optionalDependencies` 😓 ideally they should be `devDependencies` as they are only needed for vscode-jest developers, but yarn install failed on our minimal node 8 env because some dependencies require higher node versions, even though they are `devDependencies`. When yarn sees integrity issue on the  `optionalDependencies`it would just give warnings and move on. I could turn off the whole integrity checking (--ignore-engine) but I figure it is better to at least verify the integrity of the regular dependencies. 
- updated the travis config to split the tasks into different stages/jobs so we can give different node versions to test.
- while in there, I also added a few more node versions and os, including windows.

now the PR check definitely took longer but hopefully it is worthy as it verifies more environments...

---

I don't know why the coverage dropped so dramatically, most of the files dropped are not even changed, not sure if it is due to the different version brings in different calculation... if you guys are ok with it, I am just going to ignore those 😅 ...

stephtr:
VS Code is using Node >= 10 since quite some time, I therefore would be totally fine with increasing our requirements.

Concerning the split between `dependencies` and `devDependencies`: Since we are using webpack, VS Code doesn't have to install any dependencies when installing our extension. With moving some dependencies back to `dependencies`, I fear that this could increase installation time, download size and our disk footprint.

> I don't know why the coverage dropped so dramatically, most of the files dropped are not even changed, not sure if it is due to the different version brings in different calculation...

Either that or due to the deleted typings file or it also tracks the number of lines also for the travis config. Anyway, doesn't matter.

connectdotz:
> VS Code is using Node >= 10 since quite some time (at least February 2019), I therefore would be totally fine with increasing our requirements.

great, I couldn't find where vscode documented its minimal nodejs version, but I completely agree if vscode itself won't run on node 8 then there is no point to support it here. I have updated the travis accordingly, but just to be super sure, do you know where we can verify it?

> Since we are using webpack, VS Code doesn't have to install any dependencies when installing our extension. With moving some dependencies back to dependencies, I fear that this could increase installation time, download size and our disk footprint.

Good point, but based on vscode [publishing with webpack](https://code.visualstudio.com/api/working-with-extensions/bundling-extension#publishing), this is done via adding node_modules to `.vscodeignore` file. Looks like that was missing and we had been using the devDependencies hack to get away with it... I updated `.vscodeignore` and we should be able to use the correct dependencies declaration now. 

---

ok, now travis build passed without any hack, our dependencies are current, yarn.lock is super clean, and we cleaned up a few past hacks...  I build the `.vsix`, installed and run it with my workspaces, they looked fine, I am reasonably happy with this PR now 😄, thanks for the discussion 👍 . 

When you guys get a chance, please review the rest so we can get the v4 baseline in ASAP. thanks!

src/StatusBar.ts
```diff
@@ -17,12 +17,8 @@ interface StatusUpdateRequest {
   modes?: Mode[]
 }
 
-interface SpinnableStatusBarItem {
+interface SpinnableStatusBarItem extends Pick<vscode.StatusBarItem, 'command' | 'text' | 'tooltip'> {
```
seanpoulter, line 20: Well that's neat.

package.json
```diff
@@ -238,37 +238,37 @@
...
     "prettier-write": "yarn prettier --parser typescript --no-semi --single-quote --trailing-comma es5 --write --print-width 120",
     "prettier-project": "yarn prettier-write -- '?(__mocks__|src|tests)/**/*.ts'"
   },
-  "devDependencies": {
-    "@types/jest": "^24.0.11",
-    "@types/node": "^8.0.31",
-    "coveralls": "^3.0.0",
-    "danger": "^2",
-    "glob": "^7.1.1",
-    "husky": "^0.14.3",
```
seanpoulter, line 254: We don't need `husky` anymore?

connectdotz: I can't find any reference of it in our code base... do you recall why it was added?

seanpoulter: It was added here:
https://github.com/jest-community/vscode-jest/commit/99154c39cf458cac8b6356a1b6f401c42a1ff34a#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R114

I'm thinking it might come along with `lint-staged`. From [their docs](https://github.com/okonet/lint-staged#installation-and-setup):

> The fastest way to start using lint-staged is to run following command in your terminal:
> 
> ```
> npx mrm lint-staged
> ```
> **It will install and configure [husky](https://github.com/typicode/husky)** and lint-staged depending on code quality tools from package.json dependencies so please make sure you install (npm install --save-dev) and configure all code quality tools like Prettier, ESlint prior that.

connectdotz: looks like git command used its own husky and able to execute the "precommit" in package.json without the local husky installation... maybe something has changed after initial husky PR, almost 3 years ago? Anyway, given everything is working as expected, I think it's probably safe to remove husky.

However I did discover our lint-staged config and prettier commands are outdated and not ideal, so took the chance to clean them up as well.

connectdotz:
hi, if nobody has any objection, I will merge it in tomorrow

stephtr:
Sorry for the delay and thanks for finishing the work!

>I have updated the travis accordingly, but just to be super sure, do you know where we can verify it?

The update to Electron v10 [was part of VS Code v1.31](https://code.visualstudio.com/updates/v1_31#_nodejs-update).

> Good point, but based on vscode publishing with webpack, this is done via adding node_modules to .vscodeignore file. Looks like that was missing and we had been using the devDependencies hack to get away with it... I updated .vscodeignore and we should be able to safely use the correct dependencies declaration now.

Ah thanks, since I never looked at it in detail, I wasn't aware that the dependencies are usually just packaged into the vsix file instead of being installed from the package.json file.

Issue #576, opened 2020-04-28
v4.0 plan


connectdotz:
Moving toward the next major (awesome) version:

# enhancement
- [x] better start-up setting config. This is one of the major pain points. Moving toward a more interactive/wizard-like interface to set up settings like `jest.pathToConfig` with users' input. (#571, #244, #565, #556, #618 too many to list...) => #590 => #655
    - [x] jest-community/jest-editor-support#44 - Rename `pathToJest` 
    - [x] added new jest setting `jest.jestCommandLine` into code base. (#644)
- [x] better test matching logic: #484, #570, #491, #597, #648 => #593 (available [alpha.0](https://github.com/jest-community/vscode-jest/releases/tag/v4.0.0-alpha.0))
- [x] fixed coverage issue 
  - [x] coverage map generation/merge issues (#559, #551, #524) and update really old 3rd party dependencies => #588  (available [alpha.0](https://github.com/jest-community/vscode-jest/releases/tag/v4.0.0-alpha.0))
  - [x] formaters related issues: #411, #586, #585 => #635
- [x] move test status indicators outside of the code, so it won't interfere with formatting. (#554) => #599 
- [x] support parameterized tests (#427, #612) => #649
- [x] support manual and enhanced watch mode (#613, #668, #119, #308, #176, #408, #530, #36, #574...) => #674

# technical debt
- [x] upgrade dependencies, for better security and maintainability.  => #583  (available [alpha.0](https://github.com/jest-community/vscode-jest/releases/tag/v4.0.0-alpha.0))
- [x] migrate from tslint to eslint => #583  (available [alpha.0](https://github.com/jest-community/vscode-jest/releases/tag/v4.0.0-alpha.0))
- [x] adopt ";"  :-) => #583  (available [alpha.0](https://github.com/jest-community/vscode-jest/releases/tag/v4.0.0-alpha.0))
- [x] better troubleshooting and help documentation/examples (#544) => #674, #655
- [ ] auto close staled issues/PR so we can better focus on the active ones.

# release

[v4.0.0-alpha.5](https://github.com/jest-community/vscode-jest/releases/tag/v4.0.0-alpha.5) is the latest release and we are now finally feature-complete!

----

suggestion and discussion are welcome

distante:
I am not sure if this is the correct threat to write this but I think as a feature request for v4 should be cool to have a list of the tests in with an icon in the left bar:

this bar:
[...]

PR #583, opened 2020-05-11
code base clean up 

connectdotz:
## motivation:
part of [4.0](#576) initiative

we had some inconsistent eol problem, some are LF, and the other CRLF, which made GitHub diff going crazy... And since we upgraded prettier there are new defaults, such as [Arrow Function parentheses](https://prettier.io/docs/en/options.html#arrow-function-parentheses) that made our code inconsistent as well. Therefore, I think it is best to run prettier for the whole project to bring it up to a consistent base.

## summary
- [x] run prettier --write for the whole code base
  - moved the prettier rule to separate config file and stop repeating the same rules everywhere
  - standardize `eol` (LF) for all platforms when checking into github (added `.gitattributes` ).
  - added prettier check to ci
- [x] migrate from tslint to eslint and fixed lint errors 
  - there are still many warnings I didn't have time to fix. The future development could fix those as we come across them
- [x] convert to semi-colon

## test
besides the normal unit test, I also did some simple test with my local workspaces, they seem fine.


connectdotz:
unbelievable, the eol is tricky, took me quite a few rounds to get it working (hopefully) for Travis windows env. @stephtr when you get a chance please see if there is any unexpected consequence on windows. If you have git `core.autocrlf` set to true, then you probably won't be able to run `yarn prettier-project-check` as your checkout will be converted with windows CR-LF, not sure if we should force all checkout to LF (the default for Mac and Linux)? 

Now there are so many files touched, made me wonder maybe it is more efficient to just do the rest "non-code" mass-migration (adopting semicolon and tslint=>eslint), so we get it over in one shot instead of multiple PRs/reviews... @seanpoulter @stephtr do you guys have any objection? 

BTW, I assume we are all ok switching to semicolon, but figure I should confirm again...

package.json
```diff
@@ -227,22 +267,22 @@
     ]
   },
   "lint-staged": {
-    "*.json": "yarn prettier-write --parser json",
-    "*.ts": "yarn prettier-write --parser typescript"
+    "*.json": "yarn prettier --write",
+    "*.ts": "yarn prettier --write"
```
seanpoulter, line 271: **Non-blocking suggestion**: We should look at adding `eslint` in here too like [this example](https://github.com/okonet/lint-staged#running-multiple-commands-in-a-sequence).

connectdotz: I was actually having a 2nd thought on these post-commit hooks... The fact that the code can be magically changed behind my back after commit was not a conform to me, especially during the early days when I first involved in this project... now we have ci guard on both lint and format, I almost feel we could get rid of these post-commit hooks so we know what we see in vscode is what will show up in the github PR... 

This could be one of those subjective preferences without clear right or wrong... do you guys have a strong preference?

connectdotz:
@seanpoulter thanks for taking the time to review! 👍 

> There's a few lines that are getting close to that 120 character limit I object to but it isn't worth any effort to debate code formatting.

I agree the auto-formatting sometimes looked silly. I figure developers can always change it later if it bothers them, provided prettier check passes. 

> The one thing to point out is this may affect our 4 open PRs.

there is actually only 1 "active" PR, which is small, so hopefully, merge/rebase should not be too bad...

I will merge this in, as there is another large-scale baseline change coming up: semicolon and eslint migration... then we can start to work on the actual fun part of the v4...

I didn't hear your opinions regarding semicolon, I take it as "no objection"?  In that case, I will pull the trigger in the next PR...

seanpoulter:
Sounds good for the PRs then.

I have no objection to semi-colons. I haven't heard if Babel or TypeScript have solved the ambiguity in the grammar that Douglas Crockford has complained about. It was something like:
```js
function f() {
  return
  {
    key: 'value'
  }
}

// f() => undefined
```

I'm happy to review smaller MRs if you get a chunk of work done every day or so.

connectdotz:
didn't merge as I originally proposed... thinking it might be better to put all these internal changes in one PR so other PRs/developers don't have to merge/rebase it multiple times...

> I'm happy to review smaller MRs if you get a chunk of work done every day or so.

I will commit the rest of the changes in multiple commits within this PR, so you can do progressive reviews, knowing you have seen the priors... hope that will work. @seanpoulter once again, thanks for helping and pushing forward the v4!

seanpoulter:
Can I suggest another strategy? Since you're motivated and driving these changes for v4, why don't we prioritize keeping you unblocked. I find big PRs are hard to find time to review, so I'd say get this merged.

We could use a long-lived "v4 work in progress" branch but that's also going to slow you/us down if we have any PRs go in. Again, if we're prioritizing keeping you unblocked this isn't the best approach.

What about our PRs? Most of these broad changes are (hopefully) automated. That means we should be able to revive our PRs with a process like:
* Update the PR branch to the point right before we merge this
* Checkout the code formatting and code mod configuration and version info from the HEAD of master
* Install and run the tools
* Update the PR branch to the latest (HEAD of master)

I think that subset of files looks like:
* .gitattributes
* .travis.yml
* jsconfig.json
* package.json (the part that isn't formatting)
* prettier.config.js

If that sounds good, I'll move this into a living document that we can all edit (in the Wiki?).

<br/>

What do you think?

connectdotz:
@seanpoulter sorry just saw your reply after the eslint migration commit... 

I would admit that I do not fully grasp the strategy you are suggesting...? yes large change is harder to review, but most of the changes here are sort of "mechanical", should not have much logic implication, hopefully, should be easier to glance over for the most part. 

I plan to finish this PR this weekend, after that there should be no "massive" code change, just regular smaller scale PRs probably. Knowing this, can you elaborate on your proposal? I definitely like the idea to have a shorter cycle of code change, discussion, review... how do we make that happen?

seanpoulter:
Ah, yea, more code. 🤷 

I was suggesting we should merge what's already approved and keep the PRs separate. That way it's easier for me to review what's changed, and easier for anyone else to review the next MR without having to tackle the whole thing.

connectdotz:
> Ah, yea, more code. 🤷

hahaha, that was funny, nice emoji!

Sorry that I misunderstood your point, but hey, you only have to look at them, I have to do the actual work to research and write (well, lost of copy/paste actually)... ok, seriously, I will make you a deal, do this review and I won't make you do the semi-colon review, which is pretty safe to just check-in as long as it passed all the tests. 

> That way it's easier for me to review what's changed

You can see diff between 2 commits, for example, to see the eslint migration change vs the previous commit that you have reviewed: https://github.com/jest-community/vscode-jest/compare/7e8e3a6%5E...7e8e3a6  

Migration to eslint is actually a fun process, I learned a few new things along the way, for example, some of our tests "cheated" and that they actually did not work 😂 ; It is impossible to partially mock module function declarations like in our `TestResults.ts`. There is an interesting [jest discussion](https://github.com/facebook/jest/issues/936#issuecomment-545080082) about it. I ended up converting the (public) declarations to expressions in order to properly mock them... anyway, I hope you will have as much fun reviewing the code...

connectdotz:
@seanpoulter just wondering if you get a chance to look at the last commit yet, I was waiting to hear from you before proceeding to the semi-colon...

seanpoulter:
Hey @connectdotz, no, sorry. I gave it a quick skim but haven't reviewed the non-obvious changes like the tests that you mentioned. Feel free to add it as another commit or PR against this branch?

tests/JestProcessManagement/JestProcess.test.ts
```diff
@@ -226,25 +226,22 @@ describe('JestProcess', () => {
...
     })
 
     it('do not hangs on multiple stop() calls', async () => {
       expect.assertions(1)
       const promise = jestProcess.stop()
       jestProcess.stop()
       jest.runAllTimers()
-      await promise
-      expect(promise).resolves.toBeUndefined()
+      await expect(promise).resolves.toBeUndefined()
```
seanpoulter, line 244: 👏

tests/JestExt.test.ts
```diff
@@ -225,7 +224,7 @@ describe('JestExt', () => {
 
     it('should remove the cached decorations', () => {
       sut.onDidCloseTextDocument(document)
-      expect(sut.removeCachedDecorationTypes)
+      expect(sut.removeCachedDecorationTypes).toBeCalled()
```
seanpoulter, line 227: Good find! 🕵️

tests/DebugCodeLens/DebugCodeLensProvider.test.ts
```diff
@@ -33,7 +33,7 @@ jest.mock('vscode', () => {
     end: Position
 
     constructor(start, end) {
-      rangeConstructor(...arguments)
+      rangeConstructor()
```
seanpoulter, line 36: **Question**: Should we pass the `start` and `end` here? I don't remember what I was thinking 3 years ago. 😆 I'm assuming the tests pass without it which also makes me worry. Heh.

tests/Coverage/index.test.ts
```diff
@@ -1,9 +1,10 @@
 jest.disableAutomock()
+import * as Coverage from '../../src/Coverage'
 
 describe('Coverage', () => {
   describe('module.exports', () => {
     it('should match the snapshot', () => {
-      expect(require('../../src/Coverage')).toMatchSnapshot()
+      expect(Coverage).toMatchSnapshot()
```
seanpoulter, line 7: Feel free to remove this test and the snapshot if you want. Why did I add a snapshot test of the API?! Ick.
(Also, we can also defer that.)

tests/Coverage/Formatters/GutterFormatter.test.ts
```diff
@@ -141,7 +141,7 @@ describe('GutterFormatter', () => {
       expect(actual.partiallyCovered).toEqual([])
     })
 
-    it('should reindex the line numbers', () => {
+    it('should reindex the line numbers for partially covered', () => {
```
seanpoulter, line 144: Thank you. 🙏

tests/TestResults/index.test.ts
```diff
@@ -1,9 +1,10 @@
 jest.disableAutomock()
+import * as tr from '../../src/TestResults'
 
 describe('TestResults', () => {
   describe('module.exports', () => {
     it('should match the snapshot', () => {
-      expect(require('../../src/TestResults')).toMatchSnapshot()
+      expect(tr).toMatchSnapshot()
```
seanpoulter, line 7: We can probably delete this file and the snapshots too. As much as it may help keep the internal API stable, this isn't testing the behaviour of the system. Sorry for my past mistakes. 🤮

seanpoulter: We can also do this in another PR.

tests/JestExt.test.ts
```diff
@@ -27,6 +27,8 @@ import { updateCurrentDiagnostics } from '../src/diagnostics'
 import { JestProcessManager, JestProcess } from '../src/JestProcessManagement'
 import * as messaging from '../src/messaging'
 
+/* eslint jest/expect-expect: ["error", { "assertFunctionNames": ["expect", "expectItTakesNoAction"] }] */
+
```
seanpoulter, line 31: We might be able to do this "clean"-ly using a custom matcher -- `expect.extend(fn)`,  https://jestjs.io/docs/en/expect#expectextendmatchers. I'd keep it as is for now.

.vscode/settings.json
```diff
@@ -4,16 +4,14 @@
...
     "node_modules/jest-editor-support/src": false
   },
   "typescript.tsdk": "./node_modules/typescript/lib",
-  "eslint.enable": false,
-  "tslint.autoFixOnSave": true,
-  "prettier.semi": false,
-  "prettier.printWidth": 120,
-  "prettier.trailingComma": "es5",
-  "prettier.singleQuote": true
+  // eslint
```
seanpoulter, line 12: Does VSCode complain about the comment in JSON? It's not always allowed. 🤔

tests/TestResults/TestResult.test.ts
```diff
@@ -25,40 +27,21 @@ describe('TestResult', () => {
...
-        const expected = {}
-        testResultsWithLowerCaseWindowsDriveLetters.mockReturnValueOnce(expected)
-
+        const spy = jest.spyOn(TestResult, 'testResultsWithLowerCaseWindowsDriveLetters')
         const data: any = { coverageMap: {} }
-        expect(resultsWithLowerCaseWindowsDriveLetters(data)).toEqual({
-          coverageMap: expected,
-        })
+        resultsWithLowerCaseWindowsDriveLetters(data)
+        expect(spy).toHaveBeenCalled()
```
seanpoulter, line 44: Huh. Yea, that is messy.

seanpoulter:
> @seanpoulter just wondering if you get a chance to look at the last commit yet, I was waiting to hear from you before proceeding to the semi-colon...

Done. There's no blocking comments. Nicely done.

.vscode/settings.json
connectdotz: it's true comments are not part of the official json scheme, but vscode .settings is fine with the comments as long as I can remember...

tests/DebugCodeLens/DebugCodeLensProvider.test.ts
connectdotz: see line 4, it got mocked, I didn't dig deeper to see why. When eslint is complaining about using `arguments`, I realized it is a mocked function and we don't check the arguments, so no need to pass anything, thus removed. If it is intended to test the actual constructor, then the original test is not doing that...

tests/Coverage/index.test.ts
connectdotz: cool, removed.

tests/TestResults/index.test.ts
connectdotz: removed

tests/JestExt.test.ts
connectdotz: agree, but figure it might not be worthy to do it just for one test file, would be a fun exercise for the future PR.

connectdotz:
decided to wrap prettier under eslint, so all format errors will be shown just like lint  error. Together with eslint's auto fix on save and ci check, don't think we need lint-staged during pre-commit anymore, so it is gone too...

I think the package.json looked reasonably clean now. If no objection, I am going to pull the trigger for the following format changes (prettier):
- semi: true
- printWidth: 100
  - I was thinking about @seanpoulter comment about 120 is too long and 80 is too short sometimes, so let's split in the middle then?

connectdotz:
I think I have touched enough files in this PR 😱   Ready to merge now. Last call for questions/concerns... @seanpoulter @stephtr

tests/TestResults/index.test.ts
seanpoulter: Thanks!

Issue #585, opened 2020-05-20
GutterFormatter images are not in build


markcarroll:
### Environment

1. `node -v`: v12.16.1
1. `npm -v`: 6.14.5
1. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): jest@26.0.1
1. your vscode-jest settings if customized:
    - jest.pathToJest? default
    - jest.pathToConfig? default
    - anything else that you think might be relevant? see below

1. Operating system: macOS 10.15.4

### Prerequisite
- are you able to run jest test from command line? yes
- how do you run your tests from command line? (for example: `npm run test` or `node_modules/.bin/jest`) jest (with npx fallback)

### Steps to Reproduce

Turn on GutterFormatter in vscode settings

### Relevant Debug Info

VSCode fails to load the SVG images used in the gutter formatter. It is trying to load them from:

> Request URL: file:///Users/mark/.vscode/extensions/orta.vscode-jest-3.2.0/src/Coverage/Formatters/GutterFormatter/partially-covered-gutter-icon.svg

That folder is not present in extension folder. 

### Expected Behavior

Gutter images load. 

### Actual Behavior

Error as listed above. 

--- 

The fastest (and the most fun) way to resolve the issue is to submit a pull-request yourself. If you are interested, feel free to check out the [contribution guide](https://github.com/jest-community/vscode-jest/CONTRIBUTING.md), we look forward to seeing your PR...

Tymek:
I'll take a shot at it. My first guess is that `raw-loader` or something like that can help. We'll see

PR #598, opened 2020-06-26
Status indicator in gutter

### LINKED ISSUE #554: [style] Indentation problem with status indicator
Is it possible to move the status indicator to above the test instead of having it prefixed? I don't like how it messes with the indentation. 
Before when a dot was used as the indicator the indentation worked fine.

image

Tymek:
closes #554

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Bug-fixes within the same version aren't needed
 * upgrade all dependencies to the latest, except istanbul-lib-xxx, which requires more code change and will be handled in a separate coverage PR. - @connectdotz
 * code base clean up: migrate from tslint to eslint and adopted the latest recommended coding style, adopt semi-colon, added more ci check... - @connectdotz
 * resolve coverage map merge issue, upgrade istanbul dependencies to the latest and move to async coverageMap update. @connectdotz
+* introducing a new matching algorithm to use context relationship in matching test blocks and assertions - @connectdotz
 -->
 
 ### 3.2.0
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,7 +1,8 @@
 module.exports = {
   preset: 'ts-jest',
   testEnvironment: 'node',
-  testRegex: 'tests/.*\\.ts$',
+  testRegex: 'tests/.*\\.test\\.ts$',
+  collectCoverageFrom: ['src/**/*.ts'],
   automock: true,
   moduleFileExtensions: ['ts', 'js', 'json'],
   unmockedModulePathPatterns: [
--- a/package.json
+++ b/package.json
@@ -279,7 +279,7 @@
   "dependencies": {
     "istanbul-lib-coverage": "^3.0.0",
     "istanbul-lib-source-maps": "^4.0.0",
-    "jest-editor-support": "^27.2.0",
+    "jest-editor-support": "^28.0.0-beta",
     "jest-snapshot": "^25.5.0"
   },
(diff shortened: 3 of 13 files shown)


src/decorations.ts
```diff
@@ -1,86 +1,41 @@
...
-export function failingItName() {
+function createTestStateDecoration(
+  context: ExtensionContext,
+  icon: string,
+  overviewRulerColor: string
+): TextEditorDecorationType {
   return window.createTextEditorDecorationType({
-    overviewRulerColor: 'red',
+    overviewRulerColor,
+    gutterIconPath: context.asAbsolutePath(`images/icons/${icon}.svg`),
```
connectdotz, line 16: will this path be a problem in windows? how about this

```suggestion
    gutterIconPath: context.asAbsolutePath(path.join('images', 'icons', `${icon}.svg`)),
```

src/decorations.ts
```diff
@@ -1,86 +1,41 @@
...
+import {
+  window,
+  OverviewRulerLane,
+  DecorationRangeBehavior,
+  ExtensionContext,
+  TextEditorDecorationType,
+} from 'vscode';
 
-export function failingItName() {
+function createTestStateDecoration(
```
connectdotz, line 9: good move to consolidate the functions. 👍

src/decorations.ts
Tymek: tested with on Win10+VSCodium, clean install. It does this internally I think

src/decorations.ts
Tymek: thx

Tymek:
1. changelog updated. tests done
2. can do on monday. using icons from [vscode-icons](/microsoft/vscode-icons). anyway, when will next release happen? I might be able to add customizable colors and icon sets

connectdotz:
> when will next release happen?

I am thinking about cutting a beta soon... if you can get the new icons in on Monday, I will wait.

>  I might be able to add customizable colors and icon sets

hmmm... let's hold off on that, I think the default set that works in both lite/dark mode should be sufficient.

Tymek:
I think selected colors are readable on both dark and light backgrounds. It's not worse than before, so should be enough for now. Customization will make dark/light versions obsolete.

SVGs from `vscode-codicons` under MIT license (https://git.io/JJv25)

I opted to use square for skipped tests and circle for not-ran. It looks much better this way. Counterargument could be that Jest itself is using empty circle for skipped tests, so it's not super-consistent.

GIF in Readme needs an update for 4.0 :slightly_smiling_face:

src/decorations.ts
Tymek: changed to what is already in repo (see [src/Coverage/Formatters/GutterFormatter/index.ts:46](/jest-community/vscode-jest/blob/d84d24f9f2f39db43e327bf896e706e43ac4f10e/src/Coverage/Formatters/GutterFormatter/index.ts#L46))

connectdotz: actually there is a bug regarding GutterFormatter icon (#585), but it might be a packaging issue and not necessarily the path... I don't have any windows machine, not sure if there is any variation that could cause a path problem... unless we are absolutely sure otherwise it is probably better to be safe... 

I actually think we should update GutterFormatter 1. the path formation, 2. consolidate the svg files to the same location as yours.

Tymek:
devops issues. I will open new pull request with the same changes but from new branch (and remember not to make the same mistake again, when pull bot wiped my master). Link should appear below in a sec

PR #638, opened 2020-12-14
Add JSON schema support for Jest configuration files and package.json

igetgames:
Provides tooltips, autocomplete, and validation of Jest configuration files. 

`package.json`:

image

`jest.*json`:

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,16 @@
         "path": "./syntaxes/jest-snapshot.tmLanguage"
       }
     ],
+    "jsonValidation": [
+      {
+        "fileMatch": "jest.*json",
+        "url": "https://askirmas.github.io/jest.schema.json"
+      },
+      {
+        "fileMatch": "package.json",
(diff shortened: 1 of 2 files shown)


connectdotz:
@igetgames thanks for the PR, I think this is a great idea 👍 .

However, I noticed there is an existing PR in jest (facebook/jest#9879) that tries to correct and pull in the schema. We should probably wait when the schema is properly vetted and preferably merged into the jest official repo. Let's keep an eye on that PR...

igetgames:
Makes sense @connectdotz, thanks for the heads up. I had come across @askirmas's site while looking for the schema, not realizing he was trying to get it into Jest directly.

askirmas:
If I remember well, guys wanted to see schema fully integrated with their TS/Docs.md/CLI. Too many stuff for me alone, so I’m just pasting schema URL in vscode settings across my projects, updating on demand and pasting snippets as `"examples"`

PR #659, opened 2021-02-04
Support for todo tests

pmcelhaney:
This is an attempt to support "todo" tests. The unit tests pass, but I must be missing something, because when I run the extension, it doesn't add the todo tag to any test.

Depends on https://github.com/jest-community/jest-editor-support/pull/61


src/Coverage/CoverageMapProvider.ts
```diff
@@ -41,7 +36,7 @@ export class CoverageMapProvider {
   setFileCoverage(filePath: string, map: CoverageMap): void {
     this._map.data[filePath] = map.fileCoverageFor(filePath);
   }
-  public getFileCoverage(filePath: string): FileCoverage {
+  public getFileCoverage(filePath: string): any {
```
pmcelhaney, line 39: TypeScript was complaining about FileCoverage not having a bunch of properties. I couldn't figure out how to add the missing properties, so I fudged it.

package.json
```diff
@@ -297,7 +297,7 @@
     "vscode-codicons": "^0.0.4"
   },
   "devDependencies": {
-    "@types/istanbul-lib-coverage": "^2.0.2",
+    "@types/istanbul-lib-coverage": "^2.0.3",
```
pmcelhaney, line 300: See comment below about FileCoverage. This change was part of my attempt to address the issue. It didn't help.

src/Coverage/CoverageMapProvider.ts
connectdotz: you can add a type check, something like this:
```
public getFileCoverage(filePath: string): FileCoverage {
    const coverage = this._map.data[filePath];
    if(!(coverage instanceof FileCoverage)){
      throw new Error(`expect a FileCoverage but got ${JSON.stringify(coverage)}`);
    }
    return coverage;
  }
```
we actually only save FileCoverage in the _map, but adding a type-check is better foolproof than just casing.

connectdotz:
it didn't seem this PR is complete, in any case, it's probably better to wait after merging #662.

> but I must be missing something, because when I run the extension, it doesn't add the todo tag to any test.

are you linking with your local `jest-editor-support` package?

pmcelhaney:
Yes, I have it linked to the other package. Most of the changes in this PR were required due to type changes in the other package. Whatever is missing may be obvious to someone more familiar with TypeScript.

connectdotz:
Forgot to mention that you would also need to do `yarn prepublish` on jest-editor-support, if you haven't...

pmcelhaney:
Ah, yes, I did that once. Can't remember if I had master or the branch checked out at the time, so that may be part of the issue. 

Now that I have it set up I'm going to run the latest code instead of the alpha release in my day to day development and try to get more actively involved. 

Thanks for building this awesome productivity-enhancing tool!

connectdotz:
this PR is outdated and the todo item is now somewhat supported with TestExplore, especially after 4.2. Thanks for the contribution though.

PR #707, opened 2021-05-19
handle windows path for debugging

### LINKED ISSUE #694: v4 migration failed to debug tests correctly
### Environment

1. `vscode-jest version`: v4.0
2. `node -v`: v14.15.
3. `npm -v`: 6.14.10
4. `npm ls jest`: 26.6.3
5. your vscode-jest settings if customized:
```
  "jest.pathToJest": "./node_modules/.bin/jest",
  "jest.debugCodeLens.showWhenTestStateIn": ["fail", "unknown", "pass", "skip"]
```

6. Operating system: Windows 10

### Prerequisite
- are you able to run jest test from the command line? Yes
- how do you run your tests from the command line? npm run test (which just invokes `jest`)

### Steps to Reproduce
In a NodeJS express Typescript project, using v3.2.0, without a modified `launch.json`, and with the following vscode-jest settings:
```
  "jest.pathToJest": "./node_modules/.bin/jest",
  "jest.debugCodeLens.showWhenTestStateIn": ["fail", "unknown", "pass", "skip"]
```
and the following `jest.config.js`:
```
module.exports = {
  preset: 'ts-jest',
  testEnvironment: 'node',
  testPathIgnorePatterns: ['node_modules/', 'build/', '__tests__/utils/*'],
  rootDir: 'src',
  moduleDirectories: ['node_modules', 'src'],
  collectCoverageFrom: ['**/*.{ts,tsx}', '!**/node_modules/**', '!**/vendor/**'],
}
```
worked. I could debug, the auto run/watch worked perfectly. After today's update to v4.0.0, I couldn't debug anymore and the Code Lens stopped working, and after about 20 minutes of messing with it to update to new settings, i gave up and reverted to v3.2.0 so I could continue working.

before:
image

after:
image

### Relevant Debug Info
No error/warning output in the relevant output windows

### Expected Behavior
A smooth migration between v3.2.0 settings and v4.0.0 settings

### Actual Behavior
No migration happened, and debugging wouldn't work anymore

connectdotz:
this is mainly to fix #694

- convert the lower-case  drive-letter for `vscode.URI.fsPath` on windows to upper-case and inform jest we are passing this as a path with `--runTestsByPath`
- adjust the default debug config to add `--watchAll=false` as the react-script default test script will run in the watch mode otherwise. There is no reason to run watch during debugging anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Please add your own contribution below inside the Master section
 Bug-fixes within the same version aren't needed
 
 ## Master
+* fix debug problem for windows users (#694) - @connectdotz
 -->
 
 ### 4.0.1
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "vscode-jest",
   "displayName": "Jest",
   "description": "Use Facebook's Jest With Pleasure.",
-  "version": "4.0.2-rc.1",
+  "version": "4.0.2",
   "publisher": "Orta",
   "engines": {
     "vscode": "^1.45.0"
@@ -318,8 +318,7 @@
               "program": "^\"\\${workspaceFolder}/node_modules/.bin/jest\"",
               "args": [
                 "--runInBand",
-                "--runTestsByPath",
-                "${file}"
+                "--watchAll=false"
               ],
               "cwd": "^\"\\${workspaceFolder}\"",
               "console": "integratedTerminal",
@@ -342,8 +341,7 @@
                 "test",
                 "--env=jsdom",
                 "--runInBand",
-                "--runTestsByPath",
-                "${file}"
+                "--watchAll=false"
               ],
               "cwd": "^\"\\${workspaceFolder}\"",
               "console": "integratedTerminal",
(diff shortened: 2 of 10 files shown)


stephtr (approved):
In principle this looks good to me.

What I'm just wondering: Where is the `file not found in testFiles list` error originating from?And wouldn't it make more sense to fix the comparison there instead of relying on jest to output the paths with the same capitalization as we put it in the `args`?

connectdotz:
@stephtr great question. 

> And wouldn't it make more sense to fix the comparison there instead of relying on jest to output the paths with the same capitalization as we put it in the args?

Indeed that was one of the options I considered. But giving that we will still need to covert the source files' vscode document names (jest ony reports test files), I figure it will be more consistent that we stick to "converting on input/output points and using vscode's internal representation (lower-case vscode.URI.fsPath) during the regular vscode process.". This is similar to how to find the test results that we convert the assertions' fileName to lower-case when first received test results so we could look-up on the vscode document names without conversion afterwards...

Actually, we could have avoided the upper-case conversion since the newer jest (26.6.x) could handle these paths just fine, but the older jest/CRA seem to have problem dealing with lower-case drive letters, so figure it is better to be able to support a wider range of jest versions than asking people to upgrade.

thanks for the quick review and asking good questions. 🙏

stephtr:
Thank you for this explanation!!

PR #709, opened 2021-05-21
add --watchAll=false to non-watch tasks

connectdotz:
add `--watchAll=false` to all non-watch tasks to ensure the tasks won't get stuck since some test script automatically adds `--watch` flag like the react test script.


stephtr (approved):
Just to be sure, there is no need to also add it to the `'update-snapshot'` case?
Apart from that, I'm fine with the changes!

connectdotz:
yes, `update-snapshot` just got turned into other task types([here](https://github.com/jest-community/vscode-jest/blob/a802de0ec78beb27c43a37d2802711130060d6a2/src/JestExt/process-session.ts#L92)), which are then handled in the JestProcess...

thanks for reviewing!

PR #718, opened 2021-05-25
consolidate testFiles state in TestResultProvider

connectdotz:
This is to address the test file being ignored (no test status indicator) when "list-test-files" either error out or delayed even though the test runner itself ran fine. 

This PR consolidates the "isTestFile" logic into TestResultProvider, which lookup both "list-test-files" results and the regular test run results to determine if the given file is a test file. 

---
fix 710

### DIFF
--- a/src/TestResults/TestResultProvider.ts
+++ b/src/TestResults/TestResultProvider.ts
@@ -92,6 +92,16 @@ export class TestResultProvider implements JestExtSessionAware {
     this.testSuites = {};
   }
 
+  isTestFile(fileName: string): 'yes' | 'no' | 'unknown' {
+    if (this.testFiles?.includes(fileName) || this.testSuites[fileName] != null) {
+      return 'yes';
+    }
+    if (!this.testFiles) {
+      return 'unknown';
+    }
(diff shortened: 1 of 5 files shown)


src/TestResults/TestResultProvider.ts
```diff
@@ -92,6 +92,16 @@ export class TestResultProvider implements JestExtSessionAware {
     this.testSuites = {};
   }
 
+  isTestFile(fileName: string): 'yes' | 'no' | 'unknown' {
```
stephtr, line 95: I'm not completely happy with this return type. But compared to `boolean | undefined` or an enum, this seems to be the best solution.

stephtr (approved):

src/TestResults/TestResultProvider.ts
connectdotz: yeah, I have also considered `boolean | "unknown"` but then the `if(isTestFile())` will be true for both `true or "unknown"`, so decided to just make it more explicit with a string literal type...

thanks for the quick review! 👍

PR #733, opened 2021-06-22
Fix type for contributes.configuration.properties.jest.autoRun

davidmwhynot:
Per the documentation, this should be a valid configuration:
```json
// settings.json
{
	"jest.autoRun": "off"
}
```

However, I get the following warning when using this configuration:
image

This PR should fix this issue.


connectdotz (approved):
Good catch 👍 

Feel free to update the CHANGELOG as well

connectdotz:
merged via #736

PR #736, opened 2021-06-30
make autoRun configuration schema more accurate

tommy:
This prevents VSCode from displaying a problem on the settings file for 1) the default configuration value of null, and 2) the "off" setting.

`yarn lint && yarn test && yarn vscode:prepublish` passes, and I've attached a gif showing the new messaging.

vscode-jest-schema

Hope you'll consider this small PR, and thanks for maintaining this excellent extension!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Please add your own contribution below inside the Master section
 Bug-fixes within the same version aren't needed
 
 ## Master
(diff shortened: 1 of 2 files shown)


connectdotz (approved):
@tommy thanks for the fix 👍 looks good to me. 

It is pretty similar to #733... if you can update the changelog so we won't lose contribution history that will be great!

tommy:
🤦 Haha, didn't notice someone else had made almost an identical change! I've added a changeling entry.

PR #738, opened 2021-07-04
fix a few match issues and prepare for test provider support

### LINKED ISSUE #608: Tests with double quotes in the name are skipped
### Environment

1. `node -v`: 10.22.0
1. `npm -v`: 6.14.6
1. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): jest@26.1.0
1. your vscode-jest settings if customized: Default config from vscode to jest-tests
```
{
  // Use IntelliSense to learn about possible attributes.
  // Hover to view descriptions of existing attributes.
  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
  "version": "0.2.0",
  "configurations": [
    {
      "type": "node",
      "name": "vscode-jest-tests",
      "request": "launch",
      "args": [
        "--runInBand"
      ],
      "cwd": "${workspaceFolder}",
      "console": "integratedTerminal",
      "internalConsoleOptions": "neverOpen",
      "disableOptimisticBPs": true,
      "program": "${workspaceFolder}/node_modules/jest/bin/jest"
    }
  ]
}
```

5. Operating system: Windows 10

### Prerequisite
- are you able to run jest test from command line? Yes
- how do you run your tests from command line? (for example: `npm run test` or `node_modules/.bin/jest`) `npx jest`

### Steps to Reproduce

1. Run:
```sh
npm init
npm i jest
code .
```
2. Go to 'Run' tab in vscode (Ctrl+Shift+D in Windows)
3. Click 'create a launch.json file'
4. Select 'Debug Jest tests using vscode-jest'
5. Create file 'something.js'
```js
function something() {
  return true;
}

module.exports = something;

```
6. Create file 'something.test.js'
```js
const something = require("./something.js");

describe("something", () => {
  it('something with "quotes" in name', () => {
    expect(something()).toBe(true);
  });
});
```
7. Click 'Debug' on top of the test

### Relevant Debug Info

Output to terminal:
```sh
 ${env:CI}='vscode-jest-tests'; ${env:NODE_OPTIONS}='--require C:\Users\XXX~1\AppData\Local\Temp\vscode-js-debug-bootloader.js '; ${env:VSCODE_INSPECTOR_OPTIONS}='{"inspectorIpc":"\\\\.\\pipe\\node-cdp.1716-1.sock","deferredMode":false,"waitForDebugger":"","execPath":"C:\\Program Files\\nodejs\\node.exe","onlyEntrypoint":false,"fileCallback":"C:\\Users\\XXX~1\\AppData\\Local\\Temp\\node-debug-callback-18c05939ed997a21"}'; & 'C:\Program Files\nodejs\node.exe' '.\node_modules\jest\bin\jest' '--runInBand' 'something.test.js' '--testNamePattern' 'something with "quotes" in name'
Debugger listening on ws://127.0.0.1:58715/a8a555a3-137d-47e0-99dd-6f79ff243849
For help, see: https://nodejs.org/en/docs/inspector
Debugger attached.
Test Suites: 1 skipped, 0 of 1 total
Tests:       1 skipped, 1 total     
Snapshots:   0 total
Time:        2.453 s
Ran all test suites matching /something.test.js/i with tests matching "something with quotes in name".
```

### Expected Behavior

Test will run and attach to debugger

### Actual Behavior

Test is skipped

connectdotz:
- address these match related issues:
  - when describe block has non-literal name, the match-by-context can still fail even the test names matched. (#608)
  - edge case where we could drop parameterized tests during fallback match when "ungroup" tests.

- minor changes in preparing for the up-coming vscode test API:
  - update assertion range for matched nodes
  - allow match API to accept pre-built ContainerNode as well as list of assertions

----
resubmitted to test travis/coveralls
fixes #608

### DIFF
--- a/src/TestResults/match-by-context.ts
+++ b/src/TestResults/match-by-context.ts
@@ -27,6 +27,8 @@ import {
   ROOT_NODE_NAME,
   OptionalAttributes,
   MatchEvent,
+  NodeType,
+  MatchOptions,
 } from './match-node';
 
 export const buildAssertionContainer = (
@@ -46,7 +48,10 @@ export const buildAssertionContainer = (
         new DataNode(a.title, zeroBasedLine, a, {
           fullName: a.fullName,
           isGroup: 'maybe',
-          range: { start: zeroBasedLine, end: zeroBasedLine },
+          range: {
+            start: { line: zeroBasedLine, column: 0 },
+            end: { line: zeroBasedLine, column: 0 },
+          },
         })
       );
     });
@@ -57,6 +62,7 @@ export const buildAssertionContainer = (
   return root;
 };
 
+const UnknownRange = { start: { line: -1, column: -1 }, end: { line: -1, column: -1 } };
 export const buildSourceContainer = (sourceRoot: ParsedNode): ContainerNode<ItBlock> => {
   const isDescribeBlock = (node: ParsedNode): node is DescribeBlock => node.type === 'describe';
   const isItBlock = (node: ParsedNode): node is ItBlock => node.type === 'it';
@@ -65,7 +71,10 @@ export const buildSourceContainer = (sourceRoot: ParsedNode): ContainerNode<ItBl
     const attrs = (namedNode: NamedBlock): OptionalAttributes => ({
       isGroup: namedNode.lastProperty === 'each' ? 'yes' : 'no',
       nonLiteralName: namedNode.nameType !== 'Literal',
-      range: { start: namedNode.start?.line - 1 ?? -1, end: namedNode.end?.line - 1 ?? -1 },
+      range: {
+        start: namedNode.start ? adjustLocation(namedNode.start) : UnknownRange.start,
+        end: namedNode.end ? adjustLocation(namedNode.end) : UnknownRange.end,
+      },
     });
     if (isDescribeBlock(node)) {
       container = new ContainerNode(node.name, attrs(node));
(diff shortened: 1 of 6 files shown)


connectdotz:
ok migrated to `travis-ci.com` from the deprecated `travis-ci.org` yesterday, but still not receiving the coveralls report, sent an email to coveralls support, let's see what they say...

src/TestResults/match-by-context.ts
```diff
@@ -176,11 +188,28 @@ interface FallbackMatchResult<C extends ContextType> {
...
+  ): MaybeTestResult<C> => {
+    tNode.addEvent(event);
+    aNode.addEvent(event);
+    aNode.attrs.range = tNode.attrs.range;
+
+    if (generateResult && isDataNode(tNode) && isDataNode(aNode)) {
+      return toMatchResult(tNode, aNode, event) as MaybeTestResult<C>;
+    }
+    return undefined as MaybeTestResult<C>;
+  };
```
stephtr, line 208: Would it maybe make sense to split it up into two separate handler (one calling the other) instead of doing the (still elegant) type workaround?

stephtr (commented):
I'm sorry, but I can't finish the review in time. So far, I noticed only one thing, see the comment.

connectdotz:
@stephtr thanks for the review, we had some crisis last week so could not get to it, I will address your comment shortly.

src/TestResults/match-by-context.ts
connectdotz: I think you are right, no need to do any clever type-trick, separate methods are just fine and maybe even with less mental overhead.

connectdotz:
@stephtr is there anything else you think we should address?

connectdotz:
I will merge this as the vscode test explorer seems to be ready for releasing soon, we should move on to finish the support for it...

PR #756, opened 2021-08-23
show release message once per workspace

### LINKED ISSUE #755: Stop message from appearing every time I open a workspace
How can I stop the `vscode-jest now supports the official vscode test explorer!!` message from appearing, every time I open a workspace? Shouldn't it only appear once, then never again?

connectdotz:
fix #755

### DIFF
--- a/src/extensionManager.ts
+++ b/src/extensionManager.ts
@@ -277,23 +277,28 @@ export class ExtensionManager {
     this.showReleaseMessage();
   }
   private showReleaseMessage(): void {
-    vscode.window
-      .showInformationMessage(
-        `vscode-jest now supports the official vscode test explorer!!`,
-        'Show Test Explorer',
-        'See Details'
-      )
-      .then((value) => {
-        if (value === 'Show Test Explorer') {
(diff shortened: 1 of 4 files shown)


connectdotz:
@garyking feel free to review the PR so we can push it out sooner. Thx.

src/extensionManager.ts
```diff
@@ -277,23 +277,28 @@ export class ExtensionManager {
...
+          if (value === 'Show Test Explorer') {
+            vscode.commands.executeCommand('workbench.view.testing.focus');
+          } else {
+            vscode.commands.executeCommand(
+              'vscode.open',
+              vscode.Uri.parse(
+                'https://github.com/jest-community/vscode-jest/blob/master/README.md#how-to-use-the-test-explorer'
+              )
+            );
+          }
```
Zyie, line 299: As mentioned in #755 this should exit out early if the value returned is `undefined` so that when the user dismisses the prompt it doesn't open the readme

connectdotz: indeed!

PR #773, opened 2021-09-14
Expose parse result

### LINKED ISSUE #760: Add inline button to run test
Inspired by #509. Deno extension add inline button to run every test separately:
изображение
I think the same functionality can be implemented in this extension.

connectdotz:
expose test blocks parse results to TestExplorer when match result is not available or match failed.

## motivation
- Make the TestExplorer status menu always accessible regardless of whether tests have been run. More info see [comment](https://github.com/jest-community/vscode-jest/issues/760#issuecomment-913229066). 

- enable us to retire debug codeLens in the future: to simplify the code base and reduce editor interference (#772)
---
fixes #760

[vscode-jest-4.2.0-rc.1.vsix.zip](https://github.com/jest-community/vscode-jest/files/7158435/vscode-jest-4.2.0-rc.1.vsix.zip)

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -31,7 +31,7 @@ jobs:
         run: yarn --immutable
 
   lint-danger-typecheck:
-    name: Running TypeScript compiler & ESLint
+    name: Pre-test checks
     runs-on: ubuntu-latest
     needs: prepare-yarn-cache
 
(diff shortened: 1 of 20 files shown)


connectdotz:
@stephtr looks like you are buys, I will merge this but won't deprecate debug codeLens, we can decide what to do with it later...

PR #792, opened 2021-11-04
Fix "Commands" table formatting in `README.md`

bhrutledge:
I think the `<details>` rendering is breaking it: https://github.com/jest-community/vscode-jest/blob/master/README.md


connectdotz (approved):
@bhrutledge thank you!

PR #795, opened 2021-11-12
fix table appearence in VS Code ext marketplace

jlecordier:
I suspect the missing blank lines before the tables made VS Code to show the text inline instead of rendering the tables

image


connectdotz (approved):
thanks!

PR #846, opened 2022-03-18
Activate on workspaces with `jest.config.{mjs|cjs|ts}`

felipecrs:
Like https://github.com/facebook/docusaurus

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,8 @@
   "activationEvents": [
     "workspaceContains:jest.config.js",
     "workspaceContains:jest.config.mjs",
+    "workspaceContains:jest.config.cjs",
+    "workspaceContains:jest.config.ts",
     "workspaceContains:jest.json",
     "workspaceContains:node_modules/.bin/jest",
     "workspaceContains:node_modules/react-scripts/node_modules/.bin/jest",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -32,6 +32,7 @@
   ],
   "activationEvents": [
     "workspaceContains:jest.config.js",
+    "workspaceContains:jest.config.mjs",
     "workspaceContains:jest.json",
     "workspaceContains:node_modules/.bin/jest",
     "workspaceContains:node_modules/react-scripts/node_modules/.bin/jest",
```
felipecrs, line 38: PS: I think vscode does not watch these folders. Perhaps because there is a blacklist, or something.

Araxeus:
I think that for clarity this PR should be renamed to `Fix extension not activating if jest.config is .mjs`
also technically it could also be `.ts` too

felipecrs:
> I think that for clarity this PR should be renamed to `Fix extension not activating if jest.config is .mjs` also technically it could also be `.ts` too

Added `.ts` and `.cjs`. Also fixed the PR title, I agree.

connectdotz (approved):
looks good, thanks @felipecrs

Issue #863, opened 2022-05-11
jest doc section in VS Code Testing docs


isidorn:
Hi VS Code PM here 👋 

We are currently in the process of making the testing story more discoverable in vscode.
As part of that we are writing docs for the VS Code testing framework, and we really like your testing extension and we want to have a section on it.

This is the draft PR https://github.com/microsoft/vscode-docs/pull/5332
Do you have some good docs to point us to that we could reuse? Or would you like to contribute to our PR directly?

Thanks!

isidorn:
@connectdotz awesome that you would love to help. I think the best would be a quick start guide that is VS Code specific. And for any additional configurations we would link directly to the [jest docs](https://jestjs.io/docs/getting-started). 
Not too long, and an image or two would be welcome.

Does that make sense? What do you think?

connectdotz:
Yes that makes a lot of sense. I wanted to add a quick starting guide in our README for some time, this will be a great opportunity to do so. 

I will update our repo this week and you can review to get this going, will that work?

connectdotz:
@isidorn I have released a new version (v4.6.0) with the updated README. Please let me know if you have what you needed from the [Getting Started](https://github.com/jest-community/vscode-jest#getting-started) section or if you prefer I contribute directly to your PR?

isidorn:
@connectdotz I will take the content from the Getting Started section or just point directly to it. Thank you

PR #869, opened 2022-05-23
refactor README

connectdotz:
first draft to add a quick-start section and refactor the rest README.

address #863


connectdotz:
@isidorn I end up refactoring more than I initially intended to. Feel free to focus on the "Getting Started" section and let me know if it will work for the test doc.

isidorn:
@connectdotz thanks. I think this looks good 👏 
Do you also want to contribute the same section on our side? Or would you prefer that I do it? Thanks!

PR #896, opened 2022-08-18
added option to add the folder to the ignored ones in the popup

jonnytest1:
with this there will be another option next to "Run Setup Wizard" called "Ignore Folder" which when pressed adds the corresponding workspace fodler directly to the ignored folders in the workspace config

i was thinking about , searching the complete workspace for jest files and not enable it at all or maybe a whitelist but i think its better for the user to get notified when a workspace will not be covered by jest and to consciously decide

### DIFF
--- a/src/JestExt/core.ts
+++ b/src/JestExt/core.ts
@@ -33,6 +33,7 @@ import { JestExtExplorerContext } from '../test-provider/types';
 import { JestTestProvider } from '../test-provider';
 import { JestProcessInfo } from '../JestProcessManagement';
 import { addFolderToDisabledWorkspaceFolders } from '../extensionManager';
+import { MessageAction } from '../messaging';
 
 interface RunTestPickItem extends vscode.QuickPickItem {
   id: DebugTestIdentifier;
@@ -182,12 +183,15 @@ export class JestExt {
             const msg = `${event.error}\n see troubleshooting: ${messaging.TROUBLESHOOTING_URL}`;
             this.channel.appendLine(msg);
             this.channel.show();
-            messaging.systemErrorMessage(
-              event.error,
+
+            const messageActions: Array<MessageAction> = [
               messaging.showTroubleshootingAction,
               this.setupWizardAction('cmdLine'),
-              this.setupIgnoreAction()
-            );
+            ];
+            if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 1) {
+              messageActions.push(this.setupIgnoreAction());
+            }
+            messaging.systemErrorMessage(event.error, ...messageActions);
           } else {
             this.updateStatusBar({ state: 'done' });
           }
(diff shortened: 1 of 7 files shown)


jonnytest1:
could it be you have conflicting "prettier" and "eslint - prettier" configuration ? oO

jonnytest1:
one more hint : since prettier has a rule for line endings to be LF ones i think changing 
`* text=auto ` to `* text=false` in the .gitattributes files makes more sense 🤔

src/JestExt/core.ts
```diff
@@ -144,6 +145,14 @@ export class JestExt {
     };
   }
 
+  private setupIgnoreAction(folder: string): messaging.MessageAction {
```
connectdotz, line 148: each JestExt/core corresponding to a workspace, you can find it in `this.extContext.workspace.name`. So we don't need to pass this extra argument around from the listeners to here. It should eliminate some API changes for the exit event.

src/JestExt/core.ts
```diff
@@ -176,7 +185,8 @@ export class JestExt {
             messaging.systemErrorMessage(
               event.error,
               messaging.showTroubleshootingAction,
-              this.setupWizardAction('cmdLine')
+              this.setupWizardAction('cmdLine'),
+              this.setupIgnoreAction(event.workspaceFolder)
```
connectdotz, line 189: I think we should only add the additional action for multi-root workspaces, something like
```
if (vscode.workspace.workspaceFolders && vscode.workspace.workspaceFolders.length > 1) {
...
}
```

connectdotz (changes requested):
@jonnytest1 thanks for the PR! 

I have a few comments, let me know what you think.

src/JestExt/core.ts
jonnytest1: oh i was looking into that real quick cause this
```
this.session.context.onRunEvent.fire({
        type: 'exit',
        process,
        workspaceFolder: this.session.context.workspace.name,
      }); 
```
was quite the giveaway they're somwhere close (i didnt look further into it when i didnt find a property called session ^^)

gonna refactor to use this
`this.extContext.workspace.name `

src/JestExt/core.ts
jonnytest1: 🤔 on second thought i do also have single folder workspaces that dont work with jest where disabling it with one click might also be nice

connectdotz: for single folder workspace, I think the current activation check should be sufficient; and they can simply disable the extension via vscode if it's truly not applicable. However, multi-root workspace is different:
1. we did not perform the same activation check for each multiroot workspace folder
2. users can't simply disable the extension for a certain workspace folder via vscode, thus the introduction of the option `jest.disabledWorkspaceFolders` for more granular control.

connectdotz:
Looks like the test coverage has been reduced. @jonnytest1, do you mind adding some tests for the code change? Much appreciated.

jonnytest1:
oh lol i can scroll the checks :D

connectdotz (approved):
Looks good. The test coverage dip is all right as new changes are now indeed tested. 

There are other issues such as double reporting errors but it is not related to this PR, we will address them in another PR.

PR #898, opened 2022-08-25
Fix double error reporting

connectdotz:
When jest process failed to start, users often see 2 different error/warning popups, with inconsistent formats. This PR addresses this issue:
1. consolidate where the final message was composed to apply consistent format
2. suppress `exitProcess` event in the `ListTestFileListener` and report process error via `onResult` instead.
3. furthermore, the `onResult` error will only be reported in the pop-up if `debugMode=true`, since most system errors shall be reported (again) when running the actual tests.

### DIFF
--- a/src/JestExt/core.ts
+++ b/src/JestExt/core.ts
@@ -213,9 +213,6 @@ export class JestExt {
             actions.push(this.setupIgnoreAction());
           }
           break;
-        default:
-          console.error(`invalid message action type: ${t}`);
-          break;
       }
     }
     return actions;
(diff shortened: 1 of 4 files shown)


Wxh16144:
I'm looking forward to releasing an updated version to support this fix.

connectdotz:
@Wxh16144 This is available in [v5 pre-release](https://github.com/jest-community/vscode-jest/blob/master/release-notes/release-note-v5.md), feel free to give it a spin and let us know how it works for you. Thanks.

PR #918, opened 2022-10-13
Fix autoRun table formatting

andrericardo:
In the README the table is not rendering correctly, making it quite hard to read.

image
https://github.com/jest-community/vscode-jest#autorun

After the fix:

image
https://github.com/andrericardo/vscode-jest/tree/fix-readme-autorun-table#autorun


connectdotz (approved):
great, thanks!

Issue #923, opened 2022-10-17
camelcase args are not supported in angular cli


vhdirk:
### Environment

1. `vscode-jest version`: 5.0.0
2. `node -v`: v18.10.0
3. `npm -v` or `yarn --version`: yarn 1.22.19
4. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): 
```
webapp@1.7.0-SNAPSHOT /home/dirk/devel/webapp
├─┬ @angular-builders/jest@14.0.0
│ ├─┬ jest-preset-angular@12.1.0
│ │ └── jest@28.1.3 deduped
│ └── jest@28.1.3 deduped
├─┬ jest-extended@3.0.1
│ └── jest@28.1.3 deduped
├─┬ jest-mock-console@2.0.0
│ └── jest@28.1.3 deduped
├─┬ jest-mock-extended@2.0.7
│ └── jest@28.1.3 deduped
├─┬ jest-preset-angular@12.2.0
│ └── jest@28.1.3 deduped
├── jest@28.1.3
└─┬ ts-jest@28.0.7
  └── jest@28.1.3 deduped
```
5. your vscode-jest settings if customized:
    - jest.jestCommandLine? ` yarn run test --bail --max-workers=2`
    - jest.autoRun? [fill]
    - anything else that you think might be relevant? [fill]
6. Operating system: Debian testing

### Prerequisite
- are you able to run jest test from the command line? yes
- how do you run your tests from the command line? (for example: `npm run test` or `node_modules/.bin/jest`) `yarn run test`

### Steps to Reproduce

Yarn run test is an alias for `ng test`, which, in my case, uses the jest builder from 
https://github.com/just-jeb/angular-builders.

The relevant snippet from my `angular.json` is 
```
        "test": {
          "builder": "@angular-builders/jest:run",
          "options": {
            "configPath": "src/jest.conf.js",
            "no-cache": true,
            "polyfills": ["src/polyfills.ts"],
            "tsConfig": "src/tsconfig.spec.json"
          }
        },
```

Jest supports both camelCase as dashed arguments. Angular CLI only supports dashed argument. vscode-jest uses camelcase.
While I understand this is not an issue originating from vscode-jest, the solution can be pretty simple and should not matter for those not using angular anyway. 

See https://github.com/just-jeb/angular-builders/blob/master/MIGRATION.MD#jest-builder for more info

ulrikdom:
[...]

PR #924, opened 2022-10-17
use kebab-case

vhdirk:
This PR fixes  https://github.com/jest-community/vscode-jest/issues/923.

Depends on https://github.com/jest-community/jest-editor-support/pull/91


connectdotz (commented):
hi @vhdirk Thanks for submitting this PR 👍 

In addition to the files you updated, there are a few other places that require similar treatment, such as: 
- package.json
- DebugConfigurationProvider and its tests
- process-listeners and its tests: (there is regex looking for watch flags, not sure if it will be impacted, please check when you are at it, thx)

to be consistent. we should probably also update these test cases
- setup-jest-debug.test.ts
- wizard-helper.test.ts

connectdotz:
@vhdirk looks good. 

There are some issues in the jest-community/jest-editor-support#91, let's finish that before merging this, so we can test these changes as a whole. While the changes are quite mechanical, it is significant, we should do more testing...

vhdirk:
@connectdotz Thanks for reviewing. For what it's worth, I've been using this as my daily driver for at least a week now. No issues to be seen :)

connectdotz:
@vhdirk I tested a bit with the new kebab arguments, and bumped into a problem with react-scripts. It seems react-scripts only recognized `"--watchAll=false"` but not `"--watch-all=false"`. I opened an issue facebook/create-react-app#12801 and hopefully, it will be addressed for future versions.

This means we might have to put the CI flag (in `jest-editor-support`) back (but I vaguely remember there were some other conflicts??), or switch between watchAll and watch-all selectively, which definitely sucks and not sure if we can reliably detect react/react-native... thoughts?

connectdotz:
I think an alternative solution is to just pass a flag to jest-editor-support and have it converts all jest arguments accordingly. That way, we don't need to hard-code or change anything on the vscode-jest side (maybe?), what do you think?

vhdirk:
I could also make a PR for https://github.com/facebook/create-react-app/issues/12801, but I suppose it would still mean that every single app ever made with it will also be broken?
That's quite a bummer.

I think adding a flag is pretty much the only way forward, though I don't quite understand how that would mean that we don't have to change anything in vscode-jest. I specifically ran into issues there.

vhdirk:
How do you feel about adding https://www.npmjs.com/package/change-case as a dependency? It would make the code quite a bit cleaner.

connectdotz:
I think it might be a bit overkill, given we only need to consider camel => dash/kebab.

vhdirk:
The same author does offer the library in smaller parts:
 - https://www.npmjs.com/package/param-case
 - https://www.npmjs.com/package/camel-case

If we have to support both options, I'm not very keen on littering the code with stuff like `kebabCase ? '--watch-all' : '--watchAll'`

connectdotz:
It seems this is pretty much a single line function, I did a quick a test and it seems to work:
```ts
it('convert to kebab case', () => {
  const s = 'testNamePath'.replace(/(\B)([A-Z])/gm, '-$2').toLowerCase();
  expect(s).toEqual('test-name-path');
});
```
Therefore, I think we are ok not to add a new dependency. 

Overall, I am thinking we will leave `vscode-jest` jest flags unchanged, but adding a new setting `jest.useDashedArgs` (jest's document used the term dashed args). We will pass that argument as part of the runnerWorkspace, which is constructed [here](https://github.com/jest-community/vscode-jest/blob/8037beb4ec0d3f5942b161211809f332cbd7f52c/src/JestExt/helper.ts#L94). You can follow "settings.shell" to see how we pass a setting all the way through to `jest-editor-support`.

On the `jest-editor-support` side, the `ProjectWorkspace` and the related types need to be modified to take in the new flag. And the actual argument case conversion will only need to take place in `Runner._getArgs()`, if `jest.useDashedArgs` is true, otherwise (undefined or false), do nothing, i.e. we will not do dash => camel conversion, it's a very limited single-direction conversion on the args it added or from `options.args`.

I hope it makes sense, let me know what you think.

connectdotz:
Hi @vhdirk, are you still interested in finishing this PR?

connectdotz:
close via #1034. Thanks to everyone who participated.

PR #950, opened 2022-11-20
Retire debug codelens

connectdotz:
debug has long been available in the TestExplorer context menu. In an effort to reduce redundant features and simplify the code base, we will remove the DebugCodeLens and related settings in v5.


markcarroll:
I really do not like this change at all.  Having to go to the other side of the screen (on a 32" 4k monitor) to find the debug button in the explorer is a real pain compared to just clicking above the test you are debugging.  @connectdotz can you either revert this change or add it as a config option please?

nausaf:
this is a really bad idea, very inconvenient

PR #977, opened 2022-12-27
fix: support path based auto completion for `jsx` and `tsx`

mrazauskas:
Following up #963

Thanks for path based auto completion for `jest.mock` and similar methods. Very helpful.

It works perfect in `js` and `ts` files. I just noticed that the feature is missing if a test file has `jsx` or `tsx` extension. Adding `javascriptreact` and `typescriptreact` language selectors fixed this issue locally.

### DIFF
--- a/src/language-provider.ts
+++ b/src/language-provider.ts
@@ -58,8 +58,10 @@ export class LocalFileCompletionItemProvider
 
 export function register(): vscode.Disposable[] {
   const selector = [
-    { scheme: 'file', language: 'typescript' },
     { scheme: 'file', language: 'javascript' },
+    { scheme: 'file', language: 'javascriptreact' },
+    { scheme: 'file', language: 'typescript' },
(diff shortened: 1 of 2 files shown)


connectdotz (approved):
Great catch, thanks for the PR!

mrazauskas:
Thanks for taking a look.

By the way, is there any reason why `jest.unstable_mockModule()` is excluded from the list of APIs with path autocompletion?

I use Jest and `vscode-jest` in one medium sized ESM project. `unstable_mockModule()` is the only way to create mocks in that case, so it would be nice to have autocompletion. If `unstable_mockModule()` wasn’t included simply by mistake, I can send a PR.

connectdotz:
Is this an official API yet? I don't see it in the API doc, but it did appear in https://jestjs.io/docs/ecmascript-modules#module-mocking-in-esm ... do you know what the state of this API is?

If it is going to stay, then I guess it's ok to add it to our support list. PR is welcome.

PR #1014, opened 2023-03-22
Add clearTerminal setting

jgillick:
When auto-run is disabled, it's convenient to have the terminal clear and only show the results from the most recent run.

In this PR:
* Added a `clearTerminal` boolean setting.
* Added `clear()` to `ExtOutputTerminal`
* Clears the terminal when a test run is scheduled if the setting is set to `true`.


connectdotz (commented):
@jgillick Thanks for the PR, I think this is a useful feature. Sorry about the delayed response.

A few comments:
1. let's change the setting to "autoClearTerminal".
2. This probably only worked for the interactive runs, which is fine. But if we can find a way to make it also work for "watch" mode, that will be even better. The "watch" mode doesn't go through the "schedule" call as it is initiated outside of vscode. But it did emit other signals... see the `process-isteners` code for clues.

jgillick:
Thanks for the feedback.
* I've updated the setting name to `autoClearTerminal`
* It will now also auto-clear when in watch mode. (basically anytime it logs `> Test run started`)

connectdotz (commented):
Hey, thanks for the update. The code's looking good, but we need to improve the test coverage for the new changes. After all, this is a test extension, right? We gotta walk the talk! 😄

When you get a chance, could you also update the [Customization](https://github.com/jest-community/vscode-jest#customization) section in the README.md?

Apologies for the slow response on my end. This is a super useful change, and once those updates are made, I'll be ready to merge it in. Thanks again!

src/JestExt/helper.ts
```diff
@@ -111,6 +111,7 @@ export const getExtensionResourceSettings = (
 
   return {
     jestCommandLine: getSetting<string>('jestCommandLine'),
+    autoClearTerminal: config.get<boolean>('autoClearTerminal') ?? false,
```
connectdotz, line 114: please use `getSetting<boolean>` instead

connectdotz (changes requested):
Thanks for the update. Looks like there are some merge issues. Please see the comment.

BTW, you can `yarn run ci` before check-in to ensure the tests and formats are correct before check-in.

jgillick:
Whoops, sorry about that. It should be resolved now.

tests/test-provider/test-item-data.test.ts
```diff
@@ -1399,22 +1423,6 @@ describe('test-item-data', () => {
...
-        [env.testFile, env.testBlock].forEach((t) =>
-          expect(jestRun.vscodeRun.enqueued).toHaveBeenCalledWith(t)
-        );
-
-        env.onRunEvent({ type: 'start', process });
-        expect(jestRun.vscodeRun.started).toHaveBeenCalledTimes(2);
-        [env.testFile, env.testBlock].forEach((t) =>
-          expect(jestRun.vscodeRun.started).toHaveBeenCalledWith(t)
-        );
-      });
```
connectdotz, line 1417: Is this an intentional deletion or merge error?

connectdotz (changes requested):
It seems that maybe a test is accidentally deleted?

connectdotz:
Thanks for the PR!

PR #1016, opened 2023-04-03
Fixed spelling mistake in readme

Jazzkid0:
Just a simple correction. Changed the link to the "setup tool" page so that it no longer says "setup too".

Let me know if I should contribute a different way :)


connectdotz (approved):
Thanks!

PR #1038, opened 2023-06-23
fix typos

ykray:
Fixed typos across repo.


connectdotz (approved):
Wow, so many typos, it's embarrassing... 😓

Thank you so much for this PR. Yes, we should use the code spell checker from now. 🤭 I hope you didn't have to do all these by hand...

Issue #1060, opened 2023-09-01
Clear test results on test execution?


ladariha:
### Environment

1. `vscode-jest version`: 6.0.1
2. `node -v`: 14.17.0
3. `npm -v` or `yarn --version`: yarn 1.22.19
4. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): [fill]
5. your vscode-jest settings if customized:
    - jest.autoRun? off
    - jest.debugMode: true
    - jest.monitorLongRun: off
    - jest.autoClearTerminal: true
    - jest.virtualFolders : [{name: "somename", jestCommandLine: "/home/somePath/node_modules/jest/bin/jest.js --config=... "}]
    - anything else that you think might be relevant? [fill]
6. Operating system: RHEL 9

Would it be possible to clear previous test results (or make it configurable) when I start new test execution? What I mean is the right column in TEST RESULTS bottom panel. I'd find it much easier to use if the content of the history would be cleared anytime I trigger some tests. So essentially the desired state is to have history of size 1, aka latest test run only

![image](https://github.com/jest-community/vscode-jest/assets/611555/605030ff-5bce-4a42-9036-48b1b4e1edd0)

abm0:
Is an option `"jest.autoClearOutput"` at v6.0.0 what you're looking for?

connectdotz:
This is controlled by vscode testing subsystem, cc @connor4312 for insight.

connor4312:
We could have a setting for this; if more users request it we could go that direction. An extension could also manually run the `testing.clearTestResults` command.

connectdotz:
@connor4312 good point, we can try `testing.clearTestResults` upon starting new tests. 

One potential issue though: In a multi-root workspace environment, this will mean one run from a given folder will clear out the output from other folders... is there a way to close out the output associated with the specific controllers?

connor4312:
That's right--unfortunately no command to clear a specific controller's output.

ladariha:
Having it hidden behind some settings would be great :) 

Personally I don't really care about history of test runs - in my workflow, I don't really remember what exactly was different (either in application code or test code) in previous test executions so seeing different results brings zero value

connor4312:
Yea, I think some `testing.historyEnabled` would be the right solution here. I would want to see if there's interest from more people here before taking the time to implement it (though I would take a PR)

connor4312:
sounds good 🙂

ladariha:
[...]

Issue #1051, opened 2023-07-28
No output in Test Results tab, and no coverage when coverage enabled


navalex:
### Environment

1. `vscode-jest version`: 5.2.3
2. `node -v`: 18.17.0
3. `npm -v` or `yarn --version`: 9.6.7
4. `npm ls jest` or `npm ls react-scripts` (if you haven’t ejected): jest@29.6.1
5. your vscode-jest settings if customized: default settings, just enabled coverage using vscode command
6. Operating system: Ubuntu 22.04 (KDE)

### Prerequisite
- are you able to run jest test from the command line? yes
- how do you run your tests from the command line? (for example: `npm run test` or `node_modules/.bin/jest`) using `npm run test`, which looks like: jest --coverage --watch

### Steps to Reproduce

I'm making a little project using `Next.js 13.4.4` and `Capacitor.JS ^5.0.4` (to make mobile apps). I recently installed jest following the next.js doc and jest doc. If i use my npm command, in a terminal, it is working well, but when i'm using the vscode extension, it runs tests, but I only see output in terminal view, nothing in the test results tab, and even in terminal output, there is no coverage, and in the command executed by the extension, I can see the --coverage twice. If I disable the coverage, it will still have a --coverage + a --no-coverage in the same command.

### Expected Behavior

Showing test outputs in the dedicated tab, and show coverage if enabled

### Actual Behavior

No output in test results tab, only in terminal tab, and there is no coverage, even if it's written in the command args.

connectdotz:
> but I only see output in terminal view, nothing in the test results tab

Not sure what you mean by "results tab", the "OUTPUT" tab? Anyway, the jest test run output will only appear in the terminal now. And it will not show test coverage in the file unless you enable the test coverage: please see [How to use code coverage?](https://github.com/jest-community/vscode-jest#how-to-use-code-coverage). If you always want to see test coverage, you can also consider the setting `jest.showCoverageOnLoad`

Let me know if you still have any problem seeing them.

navalex:
Hm i'm talking about the "Test Results" tabs on bottom of my VSCode. I guess it would be easier if i can get the output of a specific test when i select it in the "Tests" tab for example. That is at least the feature that i would have to make me stop use the terminal jest command manually with watch.
And for the coverage, even if I enable it in the extension settings, it doesn't change anything.
Thanks for your reactivity.

Sammyjroberts:
[...]

PR #1071, opened 2023-09-20
feat: add clearTestResults option

ladariha:
Adds option to clear test results on new test run as discussed in https://github.com/jest-community/vscode-jest/issues/1060

- README.MD updated
- unit test updated
- please note: I did not increase version in package.json - I don't know the release process, if you have it automatic in merge etc. (this also means in README.MD I guess the supported version of this new setting)

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,6 +277,7 @@ Users can use the following settings to tailor the extension for their environme
 |disabledWorkspaceFolders 💼|Disabled workspace folders names in multi-root environment|[]|`"jest.disabledWorkspaceFolders": ["package-a", "package-b"]`|
 |[autoRevealOutput](#autoRevealOutput)|Determine when to show test output|"on-run"|`"jest.autoRevealOutput": "on-exec-error"`|
 |autoClearTerminal|Clear the terminal output at the start of any new test run.|false|`"jest.autoClearTerminal": true`| >= v6.0.0 |
+|clearTestResults|Clear previous test results at the start of any new test run.|false|`"jest.clearTestResults": true`| >= v6.0.2 |
 
 #### Details
 ##### jestCommandLine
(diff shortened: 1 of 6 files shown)


connectdotz:
@ladariha Thanks for the PR. A few things come to mind:

1. we already have a setting `jest.autoClearTerminal`, not sure if we should create yet another setting to control the test results tab. Maybe we can rename the settings to `autoClearOutput` so it applies to both?
2. We are in the process of cleaning up and better organizing the settings (they are getting harder to discover). I am thinking maybe we should move this `autoClearOutput` to the new runMode (#1085), given the `revealOutput` is already there.
3. Also, it seems you only check it in the test-provider `runTests()`, which is triggered by UI/TestExplorer. But we also have background auto-runs that will not go through test-provider. You can take a look at how `autoClearTerminal` was implemented (#1014) to get some idea. 
4. There is also a separate thread about better supporting the Test Results tab (#1051). Considering these Test Results tab related issues, I am wondering if we should expand the `jest.TestExplorer` setting to add a new option `useTestResults: boolean`. If false, we will just hide the test results tab; if true, then the output content and settings should just apply to both the terminal and the test-results tab...

Anyway, as you can see there are a few options for this feature. Feel free to join the discussion (#1085) so we can discuss further.

ladariha:
I apologize for long response time. I more than happy to merge the feature with `autoClearTerminal` and also the background auto-runs issue you mentioned. Please give me a week to finish it

Regarding the point 2 and 4
- I don't really have deeper knowledge here so I cannot participate in discussions about refactoring and future improvements.

connectdotz:
The change I listed above is definitely too broad for this PR. I am working on an output overhaul PR to address them, which will include clearing output for the terminal and/or test-results panel, along with other changes. I can ping you for testing/discussion if you are interested.

PR #1075, opened 2023-09-30
feat: update test statuses as they are received

### LINKED ISSUE #1061: Show test results as soon as they occur
It would be awesome if the extension would not wait until *all* tests have run to show the status of individual tests.

This especially happens if the tests are in the same file.

ffMathy:
This fixes #1061.

Requires https://github.com/jest-community/jest-editor-support/pull/110 to be merged first.

### DIFF
--- a/src/test-provider/test-provider-helper.ts
+++ b/src/test-provider/test-provider-helper.ts
@@ -191,11 +191,8 @@ export class JestTestRun implements JestExtOutput, TestRunProtocol {
       return this.options.end();
     }
 
-    if (this.parentRun) {
-      this.parentRun.end();
-      if (isVscodeRun(this.parentRun)) {
-        this.parentRun = undefined;
-      }
+    if (this.parentRun && isVscodeRun(this.parentRun)) {
+      this.parentRun = undefined;
     }
 
     this.options?.onEnd?.();
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,6 +21,7 @@ module.exports = {
     '@typescript-eslint/ban-types': 'off',
     // customize argument ignore pattern
     'no-unused-vars': 'off',
+    '@typescript-config/strict': 'off',
     '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
   },
   overrides: [
--- a/.hintrc
+++ b/.hintrc
@@ -1,0 +1,8 @@
+{
+  "extends": [
+    "development"
+  ],
+  "hints": {
+    "typescript-config/strict": "off"
+  }
+}
--- a/src/JestExt/core.ts
+++ b/src/JestExt/core.ts
@@ -1,6 +1,4 @@
 import * as vscode from 'vscode';
-import { JestTotalResults } from 'jest-editor-support';
-
 import { statusBar, StatusBar, Mode, StatusBarUpdate, SBTestStats } from '../StatusBar';
 import {
   TestResultProvider,
@@ -44,6 +42,7 @@ import { MessageAction } from '../messaging';
 import { getExitErrorDef } from '../errors';
 import { WorkspaceManager, isInFolder } from '../workspace-manager';
 import { ansiEsc, JestOutputTerminal } from './output-terminal';
+import type { AggregatedResult } from '@jest/reporters';
 
 interface RunTestPickItem extends vscode.QuickPickItem {
   id: DebugTestIdentifier;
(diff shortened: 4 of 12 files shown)


src/test-provider/test-provider-helper.ts
```diff
@@ -191,11 +191,8 @@ export class JestTestRun implements JestExtOutput, TestRunProtocol {
       return this.options.end();
     }
 
-    if (this.parentRun) {
-      this.parentRun.end();
```
ffMathy, line 195: This line seems to break a lot of the tests. Not sure if it is a real problem though, or it can be ignored. Also not sure how to solve it.

ffMathy:
@connectdotz any chance you can help out here?

src/test-provider/test-provider-helper.ts
connectdotz: Is this change just a refactoring? I noticed the refactored logic is not the same as the original. Is there any other reason for changing the logic?

connectdotz (commented):
@ffMathy Thanks for exploring this. Sorry I was late to respond. 

This is not a complete review since it's still in draft mode, just answering your question and discussing some high level thoughts:
1. how frequently are we processing the aggregated results? If there are 10 tests, do we end up processing the first test 10 times, 2nd test 9 times etc?  Does this cause any performance issues for large projects (with lots of tests)?
2. did you get a chance to test nested describe blocks with dynamic tests (it.each for example) yet? I wonder if those complex tests will fail the partial result handling until the test structure is completed anyway. I.e. we waste a lot of resources without getting any early result updates...

src/test-provider/test-provider-helper.ts
ffMathy: > Is this change just a refactoring? I noticed the refactored logic is not the same as the original. Is there any other reason for changing the logic?

I actually added a comment on this line myself on the pull request.

It's what makes the tests fail around remembering to end the runs.

However, with this line on, when one of the tests switch status to green, the rest of them will also go out of the "running spinner" state, although they are still running.

Do you have any idea of why this happens?

ffMathy:
> @ffMathy Thanks for exploring this. Sorry I was late to respond. 

No worries at all, and thanks for the preliminary review! I'll listen to your comments and make some changes soon. Probably within a couple of weeks. 

> 1. how frequently are we processing the aggregated results? If there are 10 tests, do we end up processing the first test 10 times, 2nd test 9 times etc?  Does this cause any performance issues for large projects (with lots of tests)?

I need to investigate this. Will get back to you. I think it only updates for every test file that is run, not for each individual test. I don't think this is going to mean a lot, but I could be wrong. 

It also depends on if we can live with breaking changes in the editor support package. If not, then we have to do mappings here, which may be costly. 

> 2. did you get a chance to test nested describe blocks with dynamic tests (it.each for example) yet? I wonder if those complex tests will fail the partial result handling until the test structure is completed anyway. I.e. we waste a lot of resources without getting any early result updates...

No, not yet. Wanted to make sure I was in the right direction before doing an extensive test. I'll do that and get back to you! 

Do we have some repo with a lot of tests that can be used to try it out on? Or is it enough to use it to run the vscode-jest repo tests do you think?

src/test-provider/test-provider-helper.ts
connectdotz: > However, with this line on, when one of the tests switch status to green, the rest of them will also go out of the "running spinner" state, although they are still running.

My guess (without actually tracing the calls) is the code assumed batch mode, so once an "end" event is received, it will try to close the run. The TestExplorer will then mark all items "done" in that run accordingly. You might need to introduce a new stage/event when processing the partial result, or at least not trigger the "end" event before the whole run actually finishes...

connectdotz:
> I think it only updates for every test file that is run, not for each individual test.

If that is the case then we should be good for the context matching. I think this is the right scope for this feature.

> It also depends on if we can live with breaking changes in the editor support package. If not, then we have to do mappings here, which may be costly.

I am not sure if we want to do mapping either. Let's discuss this in jest-editor-support.

> Do we have some repo with a lot of tests that can be used to try it out on? Or is it enough to use it to run the vscode-jest repo tests do you think?

We don't really have a public testing repo, you can try with vscode-jest itself, as it has pretty complex test patterns. It's a good start if you can get vscode-jest run smoothly with the change...

BTW, I just thought about test coverage info... this will most likely need to be processed at the end of the test run. How does the reporter handle the coverage?

PR #1154, opened 2024-06-10
fix links in setup-wizard doc

pierluigigiancola:
While I was reading through the document, I found three broken links. That's all.


connectdotz (approved):
Thanks! 🙏

PR #1233, opened 2025-04-14
fix: implement just-in-time test file list updates

### LINKED ISSUE #1196: [BUG] VSCode freezes on file change in large repo
**Describe the bug**

VSCode freezes when making a file change (create / rename / delete) in a repo with lots of tests.

**To Reproduce**
1. Clone repo with many test files: `git clone https://github.com/christianvuerings/vscode-jest-large-repo.git`
2. Open VSCode with that repo: `cd vscode-jest-large-repo; code .`
3. Run all tests in test explorer (running just 1 test also repros the issue)
Image
4. After tests finish to run, rename a non-test file `src/main.js` to `src/main_renamed.js`: `mv src/main.js src/main_renamed.js`

**Expected behavior**
VSCode doesn't freeze

**Actual behavior**
VSCode freezes for 5-10 seconds

**Screenshots**

First part: shows freezing issue with 2 separate renames.
Second part: renaming works smoothly without the VSCode Jest extension

https://github.com/user-attachments/assets/12033b73-72e2-4a4f-9754-7ac703b23b2c

**Environment:**
- vscode-jest version: `v6.4.0`
- `node -v`: `v18.16.1`
- `npm -v` or `yarn --version`: `9.5.1`
- jest or react-scripts (if you haven’t ejected) version: `Jest 29.7.0`
- your vscode-jest settings:
    - jest.jestCommandLine? `None`
    - jest.runMode? `on-demand`
    - jest.outputConfig? `None`
    - anything else that you think might be relevant? Set `deferred` to `true`
- Operating system: `Mac OS 14.6.1`

**Prerequisite**
- are you able to run jest from the command line? `Yes`
- where do you run jest CLI from? `Root directory of the project`
- how do you run your tests from the command line? `npm run test`

**Additional context**

When debugging the root cause seems to be `updateTestFileList`:
https://github.com/jest-community/vscode-jest/blob/fae16e86446379f907079dab9129041869d1c585/src/JestExt/core.ts#L744-L765

Which gets called on every file create, delete or rename
https://github.com/jest-community/vscode-jest/blob/fae16e86446379f907079dab9129041869d1c585/src/JestExt/core.ts#L767-L775

If we comment out the contents of `updateTestFileList` everything works smoothly

christianvuerings:
This change optimizes file system event handling by implementing a just-in-time approach to updating test file lists. Rather than immediately refreshing the test file list on every file system event, we now mark it as dirty and only update it when necessary (before running tests).

- Add testFileListDirty flag to track when updates are needed
- Only update test file list when running or debugging tests
- Force update on initial startup for completeness
- Significantly improves performance when renaming files in large repositories

fixes #1196

### DIFF
--- a/src/JestExt/core.ts
+++ b/src/JestExt/core.ts
@@ -323,7 +323,8 @@ export class JestExt {
 
       this.events.onTestSessionStarted.fire({ ...this.extContext, session: this.processSession });
 
-      await this.updateTestFileList();
+      // Force update test file list during session initialization
+      await this.updateTestFileList(true);
 
       // update visible editors that belong to this folder
       this.updateVisibleTextEditors();
@@ -568,6 +569,11 @@ export class JestExt {
 
   //**  commands */
   public debugTests = async (debugInfo: DebugInfo): Promise<void> => {
+    // Check if test file list needs updating before debug, and update if necessary
+    if (this.testResultProvider.isTestFileListDirty()) {
+      await this.updateTestFileList();
+    }
+    
     const getDebugConfig = (
       folder?: vscode.WorkspaceFolder
     ): vscode.DebugConfiguration | undefined => {
@@ -618,6 +624,11 @@ export class JestExt {
   public async runAllTests(editor?: vscode.TextEditor): Promise<void> {
     this.enableOutputOnRun();
     await this.exitDeferMode();
+
+    // Check if test file list needs updating, and update if necessary
+    if (this.testResultProvider.isTestFileListDirty()) {
+      await this.updateTestFileList();
+    }
 
     if (!editor) {
       if (this.processSession.scheduleProcess({ type: 'all-tests', nonBlocking: true })) {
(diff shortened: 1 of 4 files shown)


connectdotz:
@christianvuerings, Thanks for the PR! A few questions before we dive into the code:

1. Does `updateTestFileList()` not cause issues in your setup? We've heard from some users that it introduces noticeable delays at startup, which is why we added the "defer" runMode. I'm wondering if it might make sense to integrate this behavior into the defer mode instead? 🤔  
2. I see you've added dirty state checks for `debug` and `runAllTests`, good, but I didn't see a similar check for individual test item runs (in `test-provider/test-item-data.ts`). Also, having to add this check in multiple places feels a bit... less than ideal. 🤔  
3. Also, should we consider letting users know when `updateTestFileList()` is running—so they understand why running a seemingly small test could have noticeable delay due to this deferred action? That said, the notice has to be non-intrusive… 
4. I am also wondering how to we let user know the test file related stats (UI) are "dirty"? For example,  in `TestResultProvider.getTestSuiteStats()`, and possibly other test-file-dependent info that we may need to test/review carefully for the impact of this new "dirty" state.

PR #1263, opened 2025-09-03
Update default for jest.runMode in table

fregante:
"undefined" should be removed from all values in the table that actually have a default.

In this case, the default value/behavior is `watch`, which I had to find elsewhere in the same document

https://github.com/jest-community/vscode-jest/blob/77d55ba477d46ac5802fb378083592173e409009/README.md#L626


connectdotz (approved):
Good point. Thanks.

